PR #146, opened 2026-06-15
Remove dead legacy VOXELPLAN parser and keep live voxel LUT support

grzanka:
This PR started as a CodeQL-driven fix for a double free in the legacy VOXELPLAN parser's SLICENO branch.

While investigating that alert, it became clear that the entire VOXELPLAN parser subtree under src/gemca/voxel was dead code: it was never wired into the build, the app-level geometry parser already rejects legacy VOX cards, and the live voxel path is the DICOM-backed DCM workflow.

This PR therefore widens scope intentionally to remove the dead legacy VOXELPLAN parser, move the still-used HU-to-WEPL selector constants into include/openshieldhit/voxel.h, and keep the active DICOM/HU LUT path intact.

Validation performed:
- focused voxel/geometry/material unit tests
- cases::05_dicom_simple
- direct run of tests/cases/05_dicom_simple


nbassler:
the entire VOXELPLAN parser was never wired into build, and since we have our own dicom reader, i think it can be revmoed entirely. However, there are som HU2LUT stuff which should migrate to proper locations.

nbassler:
this is dead code.

Issue #195, opened 2026-06-29
Checkpoint/batch scheduling & the partial-results policy knob (transport quiescence for early saves)


grzanka:
Parent: #170. Follow-up design issue raised on PR #194 by @nbassler (with @bassler). Depends-on relationship: blocks the *physical correctness* of periodic dumps in #193; informs #192; intersects #161/#168/#169/#171.

## Context — the one constraint

A scoring snapshot is only *physically complete* when every transport family spawned by the completed primaries has been drained into the scoring runtime. Today that only happens at end-of-run: the scheduler (`src/transport/osh_transport.c:85`) runs the **ion pass over all `nstat` primaries** (banking neutrons/fragments into separate pools), **then** the neutron pass, **then** `osh_simulation_run()` postprocesses and saves. A mid-pass dump at the ion safe point (`src/transport/osh_transport_ion.c:274`, `primaries_completed = primaries_done - pool->n`) therefore contains **ion-only** scoring with **zero** neutron/fragment contribution under one `completed_nstat`.

`DLET` is `Σ(LET·dose)/Σdose`; dropping the neutron/fragment numerator *and* denominator **shifts the ratio**, so a partial LET map is biased, not merely noisier. Per @nbassler/@bassler the deficit is not bounded by one wave: future families (electrons) may bank secondaries across many primary waves and process them on efficiency thresholds, so the missing fraction is scheduler-policy dependent and potentially large.

This is a **scheduling/contract** problem, not a scoring-snapshot bug. The snapshot machinery in #194 is correct and stays as-is.

## The underlying dial: the checkpoint (= batch boundary)

There is exactly one dial — *how often we bring the whole simulation to a family-complete, workers-merged, quiescent point*. Call it a **checkpoint**. A checkpoint is where (optionally) all families are drained, parallel workers sync at a barrier, per-worker private accumulators merge into the master (`osh_scoring_accumulator_merge`, #159), one variance batch folds (#169), and an (optional) dump fires. **A batch `[b, b+K)` is a `osh_worker_context` range** (`src/transport/osh_worker_context.h`) — so the checkpoint barrier and the parallel-partition barrier are the *same mechanism* (the #161 groundwork; "a periodic dump and a parallel checkpoint are the same operation", #170).

## Cost model (why this is a real trade)

[...]

Issue #193, opened 2026-06-29
Run-control C — periodic & on-demand dumps: --dump-every[-primaries]/DUMPEVERY/nsave, SIGUSR1, shadow budget reservation


grzanka:
Parent: #170 · **PR-3 of 3** · Depends on **170-A** (snapshot) + **170-B** (run-control) · Sequencing: A → B → **C**.

## Goal

Deliver the user-facing **periodic** and **on-demand** dump features end-to-end by wiring the 170-A snapshot to the 170-B run-control cadence, plus the dump CLI/beam.dat surface, `SIGUSR1`, the **shadow memory budget-reservation rule**, the headline integration test, and docs.

## Cadence + on-demand triggers

Add to `run_ctl` (struct already defined in 170-B):
```c
int run_ctl_should_dump(struct osh_run_control*, double elapsed, size_t completed);
/* true when dump_flag set, OR dump_every_s elapsed since last dump,
   OR dump_every_primaries reached since last dump; updates last_dump_* */
```
At the safe point in `run_history_range()`: `if (run_ctl_should_dump(ctl, elapsed, completed)) osh_scoring_snapshot_save(&ctl->dump_sink, live, completed, ctl->dump_outputs, ctl->dump_noutputs);`

**On-demand (P3, POSIX-only):**
```c
static volatile sig_atomic_t g_dump = 0;
static void on_sigusr1(int s){ (void)s; g_dump = 1; }   /* dump-and-continue */
```
`#if !defined(_WIN32)` → `sigaction(SIGUSR1, ...)` wired to `osh_run_control.dump_flag`. **Windows: no-op** (no SIGUSR1 equivalent); documented. Windows still gets wall-time budget, periodic time/count dumps, and graceful Ctrl-C (170-B).

## CLI + beam.dat (CLI overrides beam.dat)

- `--dump-every <dur>` and `--dump-every-primaries <N>` (`src/cli/osh_cli.{h,c}`, reusing `osh_parse_duration` from 170-B).
- beam.dat `DUMPEVERY <dur>` keyword → `beam->dump_every_s`.
- **Activate the existing dead `beam->nsave`** (the `NSTAT n step` second field, `osh_beam_parse.c:831`, logged but unused at `osh_beam.c:503`) as the default `dump_every_primaries`.

## Shadow memory: the budget-reservation rule (the cross-issue piece)

The `shadow_bytes` estimate hook lands in 170-A; the **reservation decision lives here** because it depends on the dump flags:

[...]

Issue #192, opened 2026-06-29
Run-control B — clean stop + wall-time budget: osh_run_control, safe-point drain, real completed_nstat, SIGINT/Ctrl-C


grzanka:
Parent: #170 · **PR-2 of 3** · Depends on **170-A** (snapshot primitive) · Sequencing: A → **B** → C.

## Goal

Add the **run-control transport** and the **clean-stop** primitive (P1): stop *injecting* new primaries but drain the pool so in-flight histories finish, then set `completed_nstat` to the exact count so output normalizes correctly. Wire the **wall-time budget** and **graceful SIGINT/Ctrl-C stop**. The control interface is *flags*, not signals (G3), so it stays portable.

## Run-control struct

`src/transport/osh_run_control.{c,h}`:
```c
struct osh_run_control {
    double wall_budget_s;            /* 0 = unlimited */
    double dump_every_s;             /* 0 = off  (consumed in 170-C) */
    size_t dump_every_primaries;     /* 0 = off  (consumed in 170-C) */
    volatile sig_atomic_t *stop_flag;/* NULL = never stop early */
    volatile sig_atomic_t *dump_flag;/* NULL = no on-demand dump (170-C) */
    struct osh_scoring_sink dump_sink;          /* G1 (from 170-A) */
    const size_t *dump_outputs; size_t dump_noutputs; /* G2 */
    /* internal: t_start, last_dump_s, last_dump_primaries */
};
int run_ctl_should_stop(const struct osh_run_control*, double elapsed, size_t completed);
```
`should_stop = (wall_budget_s>0 && elapsed>=wall_budget_s) || (stop_flag && *stop_flag)`.

## Transport wiring (P1 clean stop)

In `run_history_range()` (`src/transport/osh_transport_ion.c:153`), change the loop to stop injection but **drain**:
```c
int stop = run_ctl_should_stop(ctl, elapsed, primaries_completed);
while ((primaries_done < nstat && !stop) || pool->n > 0u) {
    ...
    /* at the existing safe point (after osh_particle_pool_compact, :269-291): */
    stop = run_ctl_should_stop(ctl, osh_monotonic_seconds() - t_start, primaries_completed);
}
```
[...]

Issue #170, opened 2026-06-20
Run-control: snapshot-save, periodic/on-demand dumps, wall-time budget, graceful stop


grzanka:
## Context

Follow-up to the parallelism groundwork in #161 / PR #162. This issue captures the **mid-run output-control** capability and — importantly — the generalizations needed so the *same* primitive serves native CLI, signals, and the planned WASM web target (separate issue). Four user-facing features all decompose into three primitives plus a non-destructive snapshot.

Features requested:
- **Periodic partial dump** (every N primaries or every T seconds).
- **On-demand dump** (POSIX `SIGUSR1` natively; a `postMessage` in the browser) — keep running.
- **Run by total wall-time budget** instead of (or alongside) a fixed `nstat`.
- **Graceful stop & save** on `SIGINT`/Ctrl-C (and the browser equivalent).

## Two facts in the current code that make this cheap

**1. The data model already separates requested from completed.** `osh_simulation` carries both `requested_nstat` and `completed_nstat`, and the save layer normalizes by the *completed* count:
```c
osh_simulation_save() → osh_scoring_save(ws, rt, sim->completed_nstat)
                      → inv_nstat = 1.0 / completed_nstat;   /* applied at write time */
```
`completed_nstat` is just hard-wired to `requested_nstat` today (`sim->completed_nstat = params.nstat`). Feed it the real number on early stop and normalization is correct for partial/early results for free. The ASCII/BDO writers already emit `# PRIMARIES: <count>` per file.

**2. The save layer is already per-output.** `save_one_output(ws, rt, nstat, output_idx)` writes a single output — the exact granularity needed for "dump only some pages/outputs" (used by the web target).

## The one hard constraint: postprocess is destructive

`osh_scoring_postprocess()` mutates accumulators **in place** and clears flags:
```c
page->acc.data[i] *= OSH_MEVG2GY;            /* dose: MeV/g → Gy */
page->acc.data[i] /= page->acc.data2[i];     /* LET/Qeff two-pass finalize */
page->has_data2 = 0;  page->divide = 0;
```
So a mid-run dump must **snapshot → postprocess the copy → save the copy**, never touching live state. The clone is exactly the PR-A/PR-C primitive (`osh_scoring_accumulator_alloc` + copy / `merge`).

## The safe point already exists

[...]

PR #199, opened 2026-06-30
Run-control B: clean stop + wall-time budget + SIGINT/Ctrl-C (#192)

grzanka:
## Summary

Implements **issue #192** (PR-2 of 3 under #170): the portable **run-control transport** and the **clean-stop** primitive (P1) — stop *injecting* new primaries but drain the pool so in-flight histories finish, then set `completed_nstat` to the exact count so output normalises correctly. Wires the **wall-time budget** and a **graceful SIGINT / Ctrl-C stop**.

The external stop source is a **generic callback** (`int should_stop(void *user)` + opaque `void *user`), not an OS signal or a typed flag, so transport carries no file/OS/signal knowledge and the same code serves the native CLI, a watchdog thread, a GUI button, and the planned WASM target. This mirrors the `fn + void *ctx` idiom #170 already uses for the pluggable scoring sink (G1). *(API revised from the original `volatile sig_atomic_t *stop_flag` per @nbassler's review — see 366662b.)*

### Taking #195 into account (family-exact partial results)

Issue #195's severity split calls out #192 as the *naturally fixable* case: **route the early stop through the scheduler so all families drain before the final partial save.** This PR does exactly that:

- The stop is consulted **only by the primary (ion) loop**, to halt new-primary *injection*. In-flight histories keep draining to termination.
- The family scheduler is **never short-circuited** — after the ion pass returns, the neutron/fragment passes still run over every secondary banked by the *completed* primaries.

So a partial save is **family-exact** for exactly the primaries that finished — never an ion-only fraction under a full `completed_nstat`. This keeps ratio quantities like `DLET = Σ(LET·dose)/Σdose` unbiased (the #195 concern). Verified manually on the NUCRE case: after a clean stop at 200 primaries, neutron transport (26 neutrons, 90 interactions) and Fermi break-up still ran, and every output normalised by 200.

Because each history is a pure function of its global index (`osh_rng_seed_history`), stopping early yields an unbiased Monte Carlo estimate of exactly the histories that completed.

## Public API

```c
enum osh_status osh_simulation_set_run_control(struct osh_simulation *sim,
                                               double wall_budget_s,
                                               int (*should_stop)(void *user),
                                               void *user);
```
`should_stop` is polled at the wavefront safe point (must be cheap/non-blocking); `wall_budget_s <= 0` and `should_stop == NULL` disables the policy (identical to not calling it). The library never includes `<signal.h>`.

## What changed

**New transport module**
- `src/transport/osh_run_control.{c,h}` — `struct osh_run_control` + `run_ctl_should_stop()` = `wall_budget reached` **OR** `should_stop(user)`. Dump-cadence scalars are declared for a stable shape but not consumed yet (PR-3 / #193 will add a symmetric `should_dump` callback).

**Transport / simulation wiring**
- `osh_transport_context` carries a borrowed `run_control` and reports the real `completed_primaries`. The ion wavefront loop stops injection but drains, re-evaluating the stop at the existing safe point (after compaction).
- `osh_simulation_set_run_control()` installs the policy; `osh_simulation_run()` arms the wall-budget baseline and plumbs the true completed count into `completed_nstat`. The save layer already normalises by it, so the `# PRIMARIES:` header reflects the partial count for free.

**CLI / beam.dat / signals**
- Shared `osh_parse_duration` (`src/common/osh_duration.{c,h}`): `30s` / `30m` / `1h` / bare seconds.
- `--max-time` (CLI) and `MAXTIME` (beam.dat) wall budget; **CLI overrides beam.dat**. New `beam->wall_budget_s`.
- `src/apps/osh/osh_signals.{c,h}`: POSIX `sigaction(SIGINT)` and Windows `SetConsoleCtrlHandler`, **flag-only / async-signal-safe**, plus the `osh_signals_should_stop` adapter that backs the library callback. `main.c` installs it and threads the callback down.

**Docs**
- `docs/user/command-line.md` and `getting-started.md` document `--max-time`, `MAXTIME`, Ctrl-C, and the family-exact partial-result semantics.

## Acceptance criteria (#192)

- [x] `osh_run_control` + `run_ctl_should_stop`; transport carries no file/signal knowledge.
- [x] Loop drains the pool on stop; `completed_nstat` set to the real count; output normalised by it.
- [x] `--max-time` / `MAXTIME` wired (CLI > beam.dat); duration parser.
- [x] POSIX `SIGINT` + Windows Ctrl-C graceful stop; behind `#ifdef`, no-op where unsupported.

## Tests

Unit tests are plain C and run on every CI OS (**Linux / macOS / Windows**); patch coverage is 100% per codecov.

- `test_osh_run_control.c` — budget boundary (inclusive `>=`), `should_stop` callback (constant + context-reading stubs), budget/callback OR, zero-budget-unlimited, NULL-safety.
- `test_osh_cli.c` — `osh_parse_duration` good/bad/edge inputs; `--max-time` parsing incl. the "requires a value" message.
- `test_osh_beam_setup.c` — `MAXTIME` → `beam->wall_budget_s` (suffix / bare / default / invalid).
- `test_osh_simulation.c` — **deterministic** end-to-end clean stop via an always-stop callback: `completed_nstat == one pool batch < requested`, partial save normalised by the true count, `# PRIMARIES:` header echoes it; plus a no-control run-to-completion guard.
- `test_osh_signals.c` — install + `raise(SIGINT)` (POSIX) and the `osh_signals_should_stop` adapter tracking the flag.
- `tests/cases/10_max_time` — end-to-end `--max-time` integration case.

Full suite: only the 4 pre-existing DICOM/version-tag failures remain locally (they pass on CI where fixtures and tags are present); all PR checks are green.

## Scope notes

- Periodic/on-demand dumps, the batch-aware scheduler, `--score-mem`/`--parallel`, and completeness labelling are **out of scope** — PR-3 (#193) and the broader checkpoint model (#195). This PR delivers only the graceful/wall-time stop that #195 flags as P1.

https://claude.ai/code/session_01BkuRahKBu4RRLwehP8UXMK


src/cli/osh_cli.c
```diff
@@ -268,6 +279,16 @@ static int parse_long_option(int argc, char *argv[], int *idx, struct osh_cli_op
         opt->mem_budget = value;
         return 0;
     }
+    if ((name_len == 8) && (strncmp(name, "max-time", name_len) == 0)) {
+        if (!value && !consume_option_arg(argc, argv, idx, arg, &value)) {
+            return set_err(err, err_cap, "unknown or invalid option '%s'", arg);
+        }
```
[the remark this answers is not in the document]
grzanka: Fixed in d01fcf7: a recognized option with a missing value now reports `option '--max-time' requires a value (e.g. 30s, 30m, 1h, or 500)` instead of the generic "unknown or invalid option". Added an assertion in `test_osh_cli.c` (`max_time_option`) covering the message.

---

src/apps/osh/osh_signals.c
```diff
@@ -0,0 +1,59 @@
...
+    (void) sig;
+    g_stop = 1; /* async-signal-safe: flag only */
+}
+
+void osh_signals_install_stop(void) {
+    struct sigaction sa;
+    sa.sa_handler = on_sigint;
+    sigemptyset(&sa.sa_mask);
+    sa.sa_flags = 0; /* no SA_RESTART: let blocking calls return so the flag is seen promptly */
+    (void) sigaction(SIGINT, &sa, NULL);
```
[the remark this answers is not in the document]
grzanka: Fixed in d01fcf7: `struct sigaction sa` is now `memset(&sa, 0, sizeof(sa))` before the explicit `sa_handler`/`sa_mask`/`sa_flags` assignments, so no implementation-defined field (e.g. `sa_restorer`) or padding reaches the libc wrapper uninitialised.

---

nbassler (commented):
I think the app-level SIGINT/Ctrl-C handling is in the right place, but I’m less convinced by exposing `sig_atomic_t volatile *` through the library/run-control API.

`sig_atomic_t` is fine for the narrow POSIX signal-handler pattern, but it makes the public library API signal-shaped. It also does not generalize cleanly to other stop sources: Windows console handlers, watchdog threads, GUI stop buttons, future named-pipe/TCP control, etc.

Could we make the library-level interface a generic callback instead, and keep signal/OS details entirely in the app layer? For example:

```c
enum osh_status osh_simulation_set_run_control(
    struct osh_simulation *sim,
    double wall_budget_s,
    int (*should_stop)(void *user),
    void *user);
```

Here,

```c
int (*should_stop)(void *user),
void *user
```

is the usual C pattern for a callback with caller-owned context:

- `should_stop` is the function the library calls at safe points.
- `user` is an opaque pointer passed back to that function, so the app can supply whatever state it needs without globals.

For example, the app could wire a signal-backed flag like this:

```c
static int should_stop_from_flag(void *user) {
    sig_atomic_t volatile *flag = user;
    return flag && *flag;
}

osh_simulation_set_run_control(sim, wall_budget_s, should_stop_from_flag, osh_signals_stop_flag());
```

On Windows, the app could instead back that callback with the appropriate thread-safe primitive for `SetConsoleCtrlHandler`, and a future control server could back it with named-pipe/TCP state. The simulation/transport layer would only ask “should I stop now?” at safe points, without knowing where that request came from.

grzanka:
@nbassler — agreed, and adopted in 366662b. The stop source is now a generic callback, with all signal/OS detail kept in the app layer:

```c
enum osh_status osh_simulation_set_run_control(struct osh_simulation *sim,
                                               double wall_budget_s,
                                               int (*should_stop)(void *user),
                                               void *user);
```

- `<signal.h>` is gone from the public API (`simulation.h`) and from `osh_run_control.h` — the library never names `sig_atomic_t` anymore; `run_ctl_should_stop` just rings `should_stop(user)` at the safe point.
- The app wires it via a ready-made adapter `int osh_signals_should_stop(void *user)` (in `osh_signals.c`); the raw flag accessor stays only for tests/introspection. A watchdog thread / GUI button / WASM `postMessage` can back the same callback with whatever primitive is correct for that source — exactly your point about the Windows console handler running in its own thread.
- This also matches the `fn + void *ctx` idiom #170 already uses for the pluggable scoring sink (G1), so the control and output seams are now consistent.
- Removed the unused `dump_flag`; when the on-demand dump trigger lands in PR-3 (#193) it'll get a symmetric `should_dump` callback rather than reintroducing a flag pointer.

No behavior change to `--max-time` / `MAXTIME` / Ctrl-C. Leaving this open for you to confirm/resolve.

---

src/transport/osh_transport_ion.c
```diff
@@ -147,16 +152,25 @@ static enum osh_status run_history_range(struct osh_worker_context *wctx,
     next_report_completed = progress_chunk;
     t_start = osh_monotonic_seconds();
     t_last_report = t_start;
+    /* Measure the wall budget from the run-level start (set by the driver
+     * before transport), falling back to this loop's start when uncontrolled. */
+    ctl_t_start = (ctl && ctl->t_start > 0.0) ? ctl->t_start : t_start;
```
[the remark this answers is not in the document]
grzanka: Fixed in 2f302b2: dropped the `> 0.0` guard. The driver always calls `osh_run_control_start()` before transport (whenever `transport_ctx.run_control != NULL`), so a non-NULL `ctl` always carries a valid baseline — including a legitimate monotonic `0.0`. It now uses `ctl_t_start = ctl ? ctl->t_start : t_start;`, so the budget is always measured from the run-level start; the comment was updated to state that contract.

---

tests/cases/10_max_time/README
```diff
@@ -0,0 +1,17 @@
...
+  * the family scheduler still drains every banked secondary, so the partial
+    result is family-exact for the primaries that completed;
+  * completed_nstat is the true (partial) count and every output is normalised
+    by it (the "# PRIMARIES:" header reflects it).
+
+The case asserts the end-to-end exit status (0).  Because a wall-time budget is
+intentionally non-deterministic, the exact completed count is pinned instead by
+the deterministic unit test
+test_osh_simulation::clean_stop_partial_result_is_exact, which stops via a
+pre-raised flag and checks completed_nstat == one pool batch.
```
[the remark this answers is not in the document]
grzanka: Fixed in 2f302b2: the README now says the unit test "stops via an always-stop callback" to match `test_clean_stop_partial_result_is_exact` after the callback refactor.

---

src/apps/osh/osh_beam_parse.c
```diff
@@ -95,33 +97,20 @@ struct _beam_dispatch_entry {
...
-    {OSH_BEAM_KEY_PRIMARY, _parse_primary},
-    {OSH_BEAM_KEY_RNDSEED, _parse_rndseed},
-    {OSH_BEAM_KEY_STRAGG, _parse_stragg},
-    {OSH_BEAM_KEY_TMAX0, _parse_tmax0},
-    {OSH_BEAM_KEY_TCUT0, _parse_tcut0},
-    {OSH_BEAM_KEY_USEBMOD, _parse_usebmod},
-    {OSH_BEAM_KEY_USECBEAM, _parse_usecbeam},
-    {OSH_BEAM_KEY_USEPARLEV, _parse_useparlev},
-    {NULL, NULL} /* sentinel */
+    {OSH_BEAM_KEY_APCORR, _parse_apcorr},       {OSH_BEAM_KEY_BEAMDIR, _parse_beamdir},
```
nbassler: you removed all the stubs?

nbassler: it is just two per line :)

src/apps/osh/osh_signals.c
```diff
@@ -0,0 +1,69 @@
...
+ * guarantees can be written in a signal handler and read elsewhere without a
+ * data race.
+ */
+static sig_atomic_t volatile g_stop = 0;
+
+#if defined(_WIN32)
+
+#include <windows.h>
+
+static BOOL WINAPI on_console_ctrl(DWORD ctrl_type) {
```
nbassler, line 23: Codex notes: on Windows, [osh_signals.c (line 19)](/home/bassler/Projects/openshieldhit/src/apps/osh/osh_signals.c:19) still backs SetConsoleCtrlHandler with volatile sig_atomic_t. That is app-level now, so the original API concern is fixed, but Windows console handlers run on another thread, so a Windows-specific thread-safe primitive would be cleaner if they want to make the Windows Ctrl-C guarantee precise.

nbassler: otherwise no further comments

src/apps/osh/osh_beam_parse.c
grzanka: Good catch — the stubs are all still there; adding `MAXTIME` just tipped the table over clang-format's bin-packing threshold so it reflowed into a 2-column grid. Restored to one-entry-per-line in abe95c2 by wrapping `_dispatch_table[]` in `// clang-format off` / `// clang-format on` (the same idiom already used for the key table in `osh_beam_parse_keys.h`), so it stays one-per-line regardless of clang-format version or entry count.

---

src/apps/osh/osh_signals.c
grzanka: Good point — made the Windows guarantee precise in 937bf6b. The flag's primitive now matches each platform's handler concurrency model:

- **Windows:** a `LONG` set with `InterlockedExchange(&g_stop, 1)` in the console-handler thread and read with `InterlockedCompareExchange(&g_stop, 0, 0)` (full barrier) by the poller — proper thread↔thread visibility/ordering.
- **POSIX:** unchanged `volatile sig_atomic_t` (the SIGINT handler runs in the interrupted thread, so the standard's signal↔main guarantee applies and the write stays async-signal-safe).

I also dropped the raw `osh_signals_stop_flag()` accessor — that was the thread-unsafe surface, and only the test used it — in favour of `osh_signals_reset_stop()`; the library already consumes the flag through the `osh_signals_should_stop` callback, and `<signal.h>` is now gone from the header too. Leaving the thread for you to resolve.

---

PR #203, opened 2026-06-30
Add benchmark configurations for 200 MeV protons in water with MCS is…

### LINKED ISSUE #133: SCAT: take a closer look at fluence distal edge shape
seems to match well for STRAGL 1 but not for SCAT 2 comparing OSH and SH12A.

nbassler:
…olation

- Created new test cases for 200 MeV protons in water with multiple scattering (MCS) isolation.
- Added `beam.dat`, `detect.dat`, `geo.dat`, and `mat.dat` files for both scat1 and scat2 scenarios.
- Configured scoring for primary-proton fluence and energy vs depth in `detect.dat`.
- Defined geometry and material properties in `geo.dat` and `mat.dat`.
- Updated README files to reflect the new test setups and instructions for running the benchmarks.


nbassler:
### Side finding from the `scat0` baseline: a ~0.6 mm range offset is interpolation, not physics

While validating the no-scatter baseline (`idd_water_200mev_scat0`, NUCRE off + STRAGG off, so every primary stops at an identical depth and the fluence distal edge is deterministic), OSH and SH12A do **not** stop in the same bin, even though both read the same `Water.txt` via `LOADDEDX`.

Reading the partial-fill of the transition bin (0.5 mm bins) gives the exact stopping depth:

| | range from z=0 |
|---|---|
| OSH | **26.073 cm** |
| SH12A | **26.009 cm** |
| `Water.txt` CSDA, fine log–log integral | 26.071 cm |
| `Water.txt` CSDA, coarse native-grid trapezoid | 26.030 cm |

(CSDA computed as ∫dE/(dE/dx) over the proton column, ρ = 1 g/cm³.)

So the ~0.6 mm (~0.25 %) difference is **a difference in how each code interpolates/integrates the identical dE/dx table**, not a physics discrepancy:

- OSH resamples the table onto a dense 500-point log–log grid and reproduces the faithful table integral (26.071 ≈ **26.073**).
- SH12A integrates closer to the coarse native (log-spaced) grid and lands ~0.6 mm shorter (26.030, with its transported value 26.009 a touch lower still from step discretisation).

OSH is actually the more accurate integral of the supplied table here.

This is worth recording because it **cleanly separates two effects** for #133:

1. a small **range/stopping offset** (interpolation scheme on the dE/dx table) — visible even at `MSCAT 0`;
2. the actual **distal-edge / scattering** signal — at `MSCAT 2` OSH keeps more primary fluence in the 1 cm² column and shows a narrower radial profile than SH12A, i.e. **OSH's Molière scatters too little laterally**.

The staircase distal-edge comparison makes the range offset visible (OSH's step ends one bin past SH12A's at `MSCAT 0`), and the radial profiles at the Bragg peak show the under-scattering at `MSCAT 2`.

nbassler:
image

nbassler:
image
SCAT 2 Moliere is now spot on.

nbassler:
image

nbassler:
image
SCAT 1 Gaussian is off, but that seems to be a SH12A issue, rather than OSH issue.

nbassler:
[mcs_scat_report.pdf](https://github.com/user-attachments/files/29523241/mcs_scat_report.pdf)
Full report.

grzanka:
## Code review — MCS refactor + Bethe–Molière + 200 MeV benchmarks

**Summary.** This splits the old `osh_physics_moliere` into a small MCS dispatcher (`osh_physics_scat`) with `highland` and `moliere` sibling modules, adds a clean-room full Bethe–Molière angular distribution (precomputed reduced functions + inverse-CDF table, with the magnitude anchored to the Highland width), wires mode selection through beam/transport, and adds a unit test, a Python validation tool, and the `scat0/1/2` benchmark cases. It's a clean, well-documented change overall — good module boundaries, thorough Doxygen, and the physics rationale is actually written down in `docs/physics/multiple-scattering.md` and `src/physics/README.md`.

### What reads well
- The dispatcher-plus-siblings layout is at the right altitude: dropping in Wentzel later won't touch the transport loop.
- The two scatter call sites are genuinely mutually exclusive and *say so* — physics-limited steps scatter at the hinge, boundary-limited steps scatter in phase 4 gated by `preclip_hits_boundary` (`src/transport/osh_transport_ion_step.c:757`), with a comment explaining why. `w_scat` is initialised to the incident direction in `ion_step_setup`, so the "no deflection" path is always safe.
- Precomputing the inverse-CDF table once so the hot path is an O(1) bilinear lookup is a good call, and the "no Geant3/SH12A tables (GPLv3)" clean-room note is worth having in the record.

### Worth verifying (behaviour)
1. **Mode 2 silently stops scattering as `r0 → 0` (the distal edge).** `osh_physics_moliere_scatter` derives Ω from the *macroscopic* path scale — `chic2_path ∝ path_scale_gcm2 = ctx->r0`, the residual range (`src/physics/atomic/osh_physics_scat_moliere.c:394`) — and `osh_physics_moliere_solve_b` returns 0 when `ln Ω ≤ 1.0816` (Ω ≲ 2.95). That makes `osh_physics_moliere_scatter` return 0 → `w = v`, i.e. **no deflection at all** (`osh_physics_scat_moliere.c:400-402`), whereas mode 1 in the same substep would still apply the Highland Gaussian kick of width θ₀. Since `r0` shrinks to zero at the Bragg peak, mode 2 effectively turns MCS off in the last part of the range. Given this PR is specifically about the *distal-edge fluence shape*, it's worth confirming the residual range at which Ω crosses the threshold is small enough not to bias that edge — and if it isn't, falling back to the Highland Gaussian there instead of returning 0.
2. Minor, same function: the `sin θ / θ` rejection loop (`osh_physics_scat_moliere.c:413-419`) can exhaust all 16 tries and leave `theta = 0` (→ no scatter) when `scale·ϑ` is frequently ≥ π. Rare for typical B, but it's another path that quietly drops a deflection.

### Readability
3. **`w_i` vs `wi` in `compute_moliere_constants` (`src/material/runtime/osh_material_compile.c:374-375, 395`).** Two locals that differ only by an underscore hold different quantities: `w_i` is the mass fraction, `wi` is the per-element χ_c² weight `w_i·Z(Z+1)/A`. `w_i` already means "mass fraction" in the sibling `compute_material_atomic`, so `wi` meaning something else is doubly easy to misread. Suggest e.g. `weight_i` / `chic2_weight`.
4. **`screen_z` is computed via an identity roundabout (`osh_material_compile.c:397, 408`).** `sum_w_lnz23 += wi * log(pow(z_i, 2.0/3.0))` followed by `pow(exp(sum_w_lnz23/sum_w), 1.5)` — the `2/3` and `3/2` cancel, so this is just the χ_c²-weighted geometric mean `exp(Σ wᵢ·ln Zᵢ / Σ wᵢ)`. Writing it that way states the intent and drops a `pow`+`log` per element plus a `pow`/`exp` at the end.
5. **Bare magic numbers in `osh_physics_moliere_scatter` (`osh_physics_scat_moliere.c:396, 400`):** `2.007e-5`, `3.34`, `1.167`. These are the standard Lynch–Dahl screening-angle / Ω constants, but as raw literals a reader can't map them to a formula. A named constant or a one-line `/* Lynch & Dahl … */` each would match how the rest of the file cites its sources.
6. **`d_gcm2` is essentially inert in `osh_physics_moliere_scatter`.** It's a parameter but only used in the `<= 0` guard and the `path_scale` fallback (`osh_physics_scat_moliere.c:372, 375-376`); the actual sampling runs off `path_scale_gcm2`/`chic2_coeff`. A reader reasonably expects the substep thickness to enter the distribution — a short comment ("thickness only gates/validates; magnitude comes from θ₀, shape from B") would remove the surprise.

### Minor
7. `osh_physics_mcs_scatter` copies `v → w` up front (`osh_physics_scat.c:20-22`) and then each backend copies again — harmless but redundant; the dispatcher copy only needs to cover the early-return / OFF / Wentzel paths.
8. `MSCAT 3` (Wentzel) now parses successfully (`src/apps/osh/osh_beam_parse.c`) but the run then aborts with `OSH_ENOTSUP` in `validate_transport_modes`. Accepting a value at parse time that always fails later is a small papercut — consider rejecting it at parse with a "reserved / not implemented" message, or documenting it as accepted-but-unsupported.
9. Nit: the test is `tests/unit/test_osh_physics_moliere.c` but the module is now `osh_physics_scat_moliere` (and it also exercises highland + the dispatcher); `test_osh_physics_scat*` would match the new layout. It's auto-discovered by the `test_*.c` glob, so it does build and run.

None of these block the PR; **#1** is the only one that could move physics results and it's directly on-topic for the distal edge.

---
*Reviewed by Claude (Claude Opus 4.8) on behalf of Leszek Grzanka. I reviewed the C physics/transport/material changes in depth; the `.dat`/README benchmark fixtures and `tools/plot_mcs_scat.py` got a lighter pass.*

---

Issue #169, opened 2026-06-19
Decide variance representation & parallel merge contract (Welford/West + Schubert–Gertz, batch granularity)


grzanka:
## Context

The `osh_scoring_accumulator` already scaffolds variance fields (`data_var`, `data2_var`), documented as *"running **sum-of-squares**"*, and `osh_scoring_accumulator_merge()` (PR-C, #159) already iterates them with a flat element-wise `+=`. The variance feature itself is **unwired** — nothing writes those arrays yet. That makes *now* the right moment to settle the variance representation, because the choice defines what the merge must compute, and the merge is the heart of every parallel reduction (#159) and of the partial-dump snapshot path.

Prompted by a review question citing:
- **West, D.H.D. (1979)** — *Updating Mean and Variance Estimates: An Improved Method* (incremental, weighted Welford update).
- **Schubert, E. & Gertz, M. (2018)** — *Numerically stable parallel computation of (co-)variance* (pairwise/parallel merge of two batches of unequal size).

## The decision

Pick the variance representation and, with it, the merge contract.

### Option A — additive sufficient statistics (current scaffold)
Per bin store `Σwx` (`data`) and `Σwx²` (`data_var`).
- **Merge = `+=`** — current code is already correct; `MPI_Reduce(MPI_SUM)` works directly on the raw arrays (zero custom code).
- **Con:** `Var = (Σx² − (Σx)²/N)/(N−1)` suffers **catastrophic cancellation** for high-mean / low-relative-variance bins or very large N.

### Option B — Welford/West `(n, mean, M2)` + Schubert–Gertz merge
- Numerically stable.
- **Merge ≠ add.** Two batches A, B combine by the parallel-variance formula:
  ```
  n  = nA + nB
  δ  = meanB − meanA
  mean = meanA + δ · nB / n
  M2   = M2A + M2B + δ² · (nA · nB) / n        /* the cross term a plain += cannot express */
  ```
- **Con:** MPI needs a custom `MPI_Op_create` (small, standard) instead of `MPI_SUM`. More per-bin state (`n`/`W`) and a few more FLOPs — but only at batch boundaries, not per step.

## ⚠️ Latent trap to fix regardless of choice

[...]

Issue #161, opened 2026-06-18
Tracking: Parallelism preparatory groundwork (threads / MPI / GPU)


grzanka:
## Overview

This tracking issue covers four behaviour-preserving refactoring PRs that lay the structural groundwork needed to profile and then implement parallel Monte Carlo transport (threads, MPI, GPU). No parallel code is introduced here; the goal is to establish clean interfaces so that later parallelism PRs each touch a well-bounded surface rather than requiring broad, risky surgery.

The four PRs follow a strict dependency order and can be reviewed/merged one at a time.

---

## Sub-issues (in dependency order)

All four groundwork steps landed in **#162** (merged 2026-06-25), one commit per step.

| # | Issue | Description | Status |
|---|-------|-------------|--------|
| A | #157 | Separate `osh_scoring_accumulator` from page descriptor; move `crossing_buf` to per-worker scratch | ✅ Done (#162) |
| B | #158 | Add `osh_score_deposit()` inline seam at all 11 deposit sites in `osh_scoring_step.c` | ✅ Done (#162) |
| C | #159 | Add `osh_scoring_accumulator_merge()` reduce operation with unit tests | ✅ Done (#162) |
| D | #160 | Introduce `worker_context` struct and explicit work-range history loop | ✅ Done (#162) |

> Note: `crossing_buf` relocation (part of PR-A in #157) was intentionally deferred from #162 — it requires changing the `osh_scoring_score_step()` signature, which is now tracked as part of #166.

### Follow-up work (the "honest gap" — needed before threads can be enabled)

| # | Issue | Description | Phase |
|---|-------|-------------|-------|
| | #165 | Make beam primary generation cursor-free (pass global history base explicitly) | 1a |
| | #166 | Thread per-worker accumulator set through `osh_scoring_score_step` (deposit-target redirection) | 1a |
| | #167 | Make transport profile counters per-worker (avoid data race under threads) | 1c prereq |
| | #168 | Define and CI-guard parallel reproducibility | 1c gate |
| | #169 | Decide variance representation & parallel merge contract (Welford/West + Schubert–Gertz) | — |

(#182 was a duplicate of #165 and has been closed.)

---

## Motivation

The Monte Carlo hot path (`osh_transport_ion_run_minimal` → `osh_transport_ion_step` → deposit) has two structural blockers that make any parallel scheme require global changes today:

[...]

Issue #148, opened 2026-06-16
random/transport: per-history RNG streams for order-independent, reproducible parallelism


grzanka:
> Filed by **Claude Opus 4.8** (Claude Code) on behalf of @grzanka, as the first concrete step of the SIMD / multithread / MPI / GPU parallelization effort. Scope is deliberately limited to the **RNG seeding/streaming model**; the companion scoring-reduction work is noted under *Out of scope*.

## Summary

Move the transport RNG from **two run-wide streams consumed in wavefront order** to **one independent stream per particle history, keyed by the global primary index**. The RNG *engine* (`src/random/`) is already parallel-ready; the blocker is the seeding/streaming *policy* in the transport driver, which makes the random sequence a given history sees depend on execution order (pool capacity, compaction, secondary injection). De-coupling the stream from the schedule is the prerequisite for any intra-run parallelism that is both reproducible and order-independent.

## Motivation

The parallelization roadmap (SIMD lanes, OpenMP/pthreads, MPI, GPU; see also the WASM threading notes in #16) requires that each execution lane can draw random numbers without coordinating with other lanes, and that results do not depend on *how* work was scheduled. Today neither holds for a single run. Fixing the RNG model first unblocks every backend and is a self-contained, testable refactor with no intended change to statistical results.

## Current state — what's already good (keep)

`src/random/` needs no structural change for this work:

- **No global RNG state** in the library. `struct osh_rng` (`src/random/osh_rng.h:87`) is stack-only, POD, trivially copyable — already SIMD/GPU friendly.
- **`(seed, stream)` keyed init** (`osh_rng_init`), with `stream` documented as the lane/sequence selector. Both engines accept an arbitrary 64-bit stream id: PCG32 via `inc = stream<<1|1` (`osh_rng_pcg32.c:16`), xoshiro256** via splitmix64 seeding (`osh_rng_xoshiro256ss.c:33`).
- **Threaded by pointer, not by global**: every consumer takes `struct osh_rng *rng` (`osh_transport_ion_step`, `osh_beam_new_primary`, `osh_physics_moliere_scatter`, `osh_nuclear_*`, `osh_kinematics_azimuth`). No call-graph plumbing is needed.

(The only `g_rng` global in the tree is in `examples/02_bnct/bnct_sdl.c` — not the library.)

## Problem — order-dependent stream consumption

In `osh_transport_ion_run_minimal()` (`src/transport/osh_transport_ion.c:128-129`):

```c
[...]

PR #207, opened 2026-07-01
transport: batch-aware checkpoint scheduler + partial-results policy (#195)

grzanka:
Closes part of #195.

## Context

A scoring snapshot is only *physically complete* at a **family-complete, quiescent point** — where every transport family (ions, then the neutrons/fragments they banked) spawned by the completed primaries has been drained into scoring. Observing a partial result anywhere else gives a biased ratio (e.g. `DLET = Σ(LET·dose)/Σdose` shifts when the neutron/fragment numerator *and* denominator are missing), not merely a noisier one.

Issue #195 frames this as a **scheduling/contract** problem with exactly one dial: how often the run is brought to such a checkpoint. This PR lands that contract's two foundational, non-delegated pieces — the policy data model and the batch-aware scheduler seam — with the default path byte-for-byte identical to today. The file-dump, variance, and parallel machinery that *hangs off* a checkpoint stays with its own child issues (see "Deferred").

## What this PR lands

- **`osh_checkpoint_policy`** (`src/transport/osh_checkpoint_policy.{h,c}`) — the "one dial": `mode` (final-only / live), `completeness` (exact / approx), and the time/count cadences, plus pure helpers `init`, `is_final_only`, `next_batch_size`, and `completeness_label`. Data + arithmetic only; it never touches files, clocks, threads, or scoring.
- **Batch-aware family scheduler** (`src/transport/osh_transport.c`) — an outer loop slices `[0, nstat)` into batches `[b, b+K)` and runs the family scheduler once per batch, so every checkpoint boundary is family-exact. The default (no policy / `OSH_PARTIAL_NONE`) is a single batch of `K = nstat`, i.e. the previous one-ion-pass-then-one-neutron-pass transport, unchanged.
- **Range-aware ion pass** — `osh_transport_ion_run_range([hist_lo, hist_hi))`; `osh_transport_ion_run_minimal` becomes the whole-run `[0, nstat)` wrapper.
- **Library API** — `osh_simulation_set_checkpoint_policy(sim, every_primaries)`: `0` = final-only (default, indistinguishable from not calling it); `>0` = deterministic count-cadence LIVE batches.
- **Docs** — the checkpoint/batch model, the time-vs-count cadence guidance, and completeness semantics in `docs/dev/architecture.md`.

## Why count cadence (for now)

A **count** cadence is order-independent — each history's RNG stream is a pure function of its global index — so it is deterministic and reproducible (the cadence for tests/CI). A **time** cadence self-bounds overhead independent of core count and is the production cadence; its adaptive `K ≈ rate × cadence` sizing is already implemented in `next_batch_size` but is exposed only once #193 plumbs measured throughput and periodic dumps.

## Testing

- `test_osh_checkpoint_policy` — unit coverage for every helper (final-only fast path, count/time/explicit-batch sizing, clamping, completeness label).
- `test_osh_simulation` — the final-only default completes every primary, and LIVE batches (`64,64,64,8` over `nstat=200`) transport **every** primary and match the single-batch total scored energy to `< 1e-6` relative (proving no deposits are dropped at a checkpoint).
- Full suite green (`ctest`) except a pre-existing, environment-only failure (`test_osh_main::version_components_in_string`, caused by this checkout having no semver git tag — my diff touches no version code). `bench::profile_bit_identity`, `bench::capacity_invariance`, and the `10_max_time` clean-stop case all pass, confirming the default path and the wall-time stop are unchanged.
- `clang-format` and `clang-tidy` clean; builds warning-free in Debug and Release.

## Deferred to child issues (as #195 directs)

- **#193** — CLI surface (`--dump-every`, `--dump-exact/--dump-approx`), periodic file dumps at checkpoints, the time cadence, and BDO/ASCII completeness stamping.
- **#169** — variance-batch folding at the checkpoint.
- **#161** — per-worker accumulator merge + `--score-mem`/`--parallel` at the same barrier.

This also advances the "chunked / partial run control (run in batches …)" item in `TODO.md`.

https://claude.ai/code/session_01GoAYb23sLtHYh4JjwWF6eM

---
_Generated by [Claude Code](https://claude.ai/code/session_01GoAYb23sLtHYh4JjwWF6eM)_

### DIFF
--- a/tests/unit/test_osh_simulation.c
+++ b/tests/unit/test_osh_simulation.c
@@ -482,14 +482,20 @@ static double sum_last_column(char const *path) {
 }
 
 /*
- * Drive one 00_minimal run at a given checkpoint cadence and return both the
- * completed-primary count and the total scored energy.  every_primaries == 0 is
- * FINAL-ONLY (one batch); > 0 runs family-complete batches of that size.
+ * Drive one run of test case @p case_name at a given checkpoint cadence and
+ * return the completed-primary count and total scored energy.  every_primaries
+ * == 0 is FINAL-ONLY (one batch); > 0 runs family-complete batches of that size.
+ * When @p profile_out is non-NULL, profiling is enabled and the run profile is
+ * copied out so callers can assert its counters accumulate across batches.  All
+ * bundled cases used here are water cylinders spanning z ∈ [0, 20], so the same
+ * single-quantity Energy mesh reads back the deposited energy for each.
  */
-static void run_checkpoint_case(unsigned long long nstat,
+static void run_checkpoint_case(char const *case_name,
+                                unsigned long long nstat,
                                 unsigned long long every_primaries,
                                 unsigned long long *completed_out,
-                                double *energy_sum_out) {
+                                double *energy_sum_out,
+                                struct osh_simulation_profile *profile_out) {
     char geo_path[512];
     char beam_path[512];
     char mat_path[512];
@@ -503,9 +509,9 @@ static void run_checkpoint_case(unsigned long long nstat,
     struct osh_simulation *sim = NULL;
     struct osh_results const *results = NULL;
 
-    snprintf(geo_path, sizeof(geo_path), "%s/tests/cases/00_minimal/geo.dat", OSH_PROJECT_SOURCE_DIR);
-    snprintf(beam_path, sizeof(beam_path), "%s/tests/cases/00_minimal/beam.dat", OSH_PROJECT_SOURCE_DIR);
-    snprintf(mat_path, sizeof(mat_path), "%s/tests/cases/00_minimal/mat.dat", OSH_PROJECT_SOURCE_DIR);
+    snprintf(geo_path, sizeof(geo_path), "%s/tests/cases/%s/geo.dat", OSH_PROJECT_SOURCE_DIR, case_name);
+    snprintf(beam_path, sizeof(beam_path), "%s/tests/cases/%s/beam.dat", OSH_PROJECT_SOURCE_DIR, case_name);
+    snprintf(mat_path, sizeof(mat_path), "%s/tests/cases/%s/mat.dat", OSH_PROJECT_SOURCE_DIR, case_name);
 
     /* One single-quantity TEXT mesh to a unique file, so the last column is the
      * scored energy and there is exactly one ASCII artifact to read back. */
@@ -534,10 +540,17 @@ static void run_checkpoint_case(unsigned long long nstat,
 
     ASSERT_TRUE(osh_simulation_create(beam, geo, mat, scoring, NULL, &sim) == OSH_OK);
     ASSERT_TRUE(osh_simulation_set_checkpoint_policy(sim, every_primaries) == OSH_OK);
+    if (profile_out) {
+        ASSERT_TRUE(osh_simulation_set_profiling(sim, 1) == OSH_OK);
+    }
     ASSERT_TRUE(osh_simulation_run(sim) == OSH_OK);
 
     ASSERT_TRUE(osh_simulation_get_results(sim, &results) == OSH_OK);
     *completed_out = osh_results_completed_nstat(results);
+
+    if (profile_out) {
+        ASSERT_TRUE(osh_simulation_get_profile(sim, profile_out) == OSH_OK);
+    }
 
     ASSERT_TRUE(osh_simulation_save(sim) == OSH_OK);
     *energy_sum_out = sum_last_column(out_path);
@@ -562,7 +575,7 @@ static void test_checkpoint_final_only_default_completes(void) {
 
     ASSERT_TRUE(osh_simulation_set_checkpoint_policy(NULL, 0ull) == OSH_EINVAL);
 
-    run_checkpoint_case(200ull, 0ull, &completed, &energy);
+    run_checkpoint_case("00_minimal", 200ull, 0ull, &completed, &energy, NULL);
     ASSERT_TRUE(completed == 200ull);
     ASSERT_TRUE(energy > 0.0);
 }
@@ -583,8 +596,8 @@ static void test_checkpoint_live_batches_match_final_only(void) {
     double energy_live = 0.0;
     double rel_diff;
 
-    run_checkpoint_case(200ull, 0ull, &completed_final, &energy_final); /* one batch */
-    run_checkpoint_case(200ull, 64ull, &completed_live, &energy_live);  /* 64,64,64,8 */
+    run_checkpoint_case("00_minimal", 200ull, 0ull, &completed_final, &energy_final, NULL); /* one batch */
+    run_checkpoint_case("00_minimal", 200ull, 64ull, &completed_live, &energy_live, NULL);  /* 64,64,64,8 */
 
     ASSERT_TRUE(completed_final == 200ull);
     ASSERT_TRUE(completed_live == 200ull);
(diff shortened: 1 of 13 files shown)


nbassler (commented):
src/transport/osh_transport_ion.c:332 / :334: in LIVE batching with profiling enabled, each batch overwrites prof->total_s and prof->steps with that batch’s values. Phase timers and iterations accumulate, but total wall time and steps end up representing only the last batch. That can break profiling output and benchmark accounting. I’d ask for either a per-batch local profile merged into the run profile, or an outer run-level timing/step accumulation, plus a test using osh_simulation_set_profiling() with osh_simulation_set_checkpoint_policy(..., 64).


## Test gap:
The new batching equivalence test uses 00_minimal with nuclear interactions off, so it does not really exercise the “family-complete” neutron-drain claim. I’d prefer one LIVE-vs-final test with neutron production enabled if there is a small stable fixture.

grzanka:
@nbassler Thanks — both points are correct, and I'd rather fix them here than defer.

**1. Profiling counters under LIVE batching.** You've read it exactly right. `osh_transport_ion_run_range()` is called once per batch, and its worker profile points straight at the run master (`wctx.profile = transport_ctx->profile`, zeroed once at `osh_simulation.c:372`). The per-phase timers and `iterations`/`nuclear_events`/`secondaries` accumulate into that master across batches, but the two lines at `osh_transport_ion.c:333-334`

```c
prof->total_s = total_s;
prof->steps   = (unsigned long long) steps_taken;
```

*assign* rather than accumulate, so after N batches `total_s` and `steps` carry only the last batch. That's not just noisy, it's internally contradictory — you can end up with `step_s > total_s`, which breaks the benchmark accounting.

The minimal, correct fix is to make those two accumulate like everything else:

```c
prof->total_s += total_s;
prof->steps   += (unsigned long long) steps_taken;
```

Because the master is zeroed once at profiling-enable and the default is a single batch (`K = nstat`), this is byte-identical for the final-only path (`0 + x = x`), and for LIVE batches it gives the sum of the sequential per-batch walls and steps — consistent with the phase timers. (Note this is a different axis from `osh_transport_profile_merge()`, which combines *concurrent* workers and so takes `max` on `total_s`; sequential batches genuinely sum. I'll leave a comment there so the two don't get conflated when #161 lands.)

I'll add the regression you asked for: `osh_simulation_set_profiling()` + `osh_simulation_set_checkpoint_policy(sim, 64)` over `nstat=200`, asserting `steps` equals the single-batch total and that `total_s >= step_s` (i.e. wall time is not undercounted).

**2. Test gap on the family-drain claim.** Fair — `00_minimal` runs with nuclear interactions off, so the batching-equivalence test never actually exercises a neutron drain across a checkpoint. `06_minimal_nucre` is the small nuclear-reactions fixture, so I'll add a LIVE-vs-final equivalence test on that one (every primary transported, total scored energy within tolerance) so the "family-complete, quiescent checkpoint" guarantee is covered with secondaries actually in flight.

I'll push both to the PR branch.

---

grzanka:
@nbassler Pushed `faff72e` with both fixes — and your second request turned up something worth flagging.

**1. Profiling accumulation — fixed.** `run_history_range()` runs once per checkpoint batch against the same run master, and every per-phase timer + `iterations` already accumulate into it; only `total_s`/`steps` were assigned. Changed both to `+=`. The default single batch is byte-identical (master zeroed once at profiling-enable, `0 + x = x`), and `bench::profile_bit_identity` stays green. New `test_checkpoint_profiling_accumulates_across_batches` runs a `(64,64,64,8)` cadence and asserts the LIVE step count equals the single-batch count and that the phase timers remain a consistent decomposition of the summed wall time (`phase_sum ≤ transport_s`) — the exact invariant the assignment bug violated (`step_s > total_s`).

**2. Nuclear drain test — added, and it exposed a real discrepancy.** New `test_checkpoint_live_batches_drain_families_with_nuclear` runs `06_minimal_nucre` (NUCRE on) and asserts, via the profile counters, that neutrons were actually produced and drained (`nuclear_events > 0`, `neutrons_banked > 0`) — so it genuinely exercises the family-drain path, closing the `00_minimal` gap you pointed at.

While writing it I found that **single-pass and multi-batch runs are *not* bit-identical once secondaries are in flight** — on this fixture they differ by **~2.6%** in scored energy. I chased it down:

- Not overflow: **0 neutrons dropped** in both.
- Not seeding: of the ~30 neutron-producing primaries, **all but one are bit-identical** across batchings (energies, vertex positions). A seeding defect would corrupt *every* primary past the first batch boundary.
- It's fully deterministic *within* a batching: `K=64`, `K=100`, and explicit `K=200` are each reproducible, and `K=200` matches final-only exactly. The two regimes are "one full pass" (138.21) vs "more than one batch" (141.81).

Root cause: **primary 105's ion history reaches its nuclear vertex at a fractionally different point** (`z=13.1575` vs `13.1539`) depending on whether the run is transported as one pass or split — which flips one stochastic reaction and cascades. So the RNG streams are order-independent (as the PR claims), but the *transport kernel's floating-point result* for a history is sensitive to how the run is batched. The "family-complete checkpoint ⇒ identical result" reproducibility claim doesn't hold with nuclear on.

Given that, I did **not** assert exact final-vs-LIVE equivalence — that would encode a guarantee the kernel doesn't yet provide. The test instead guards what *does* hold (completeness + per-cadence reproducibility) and documents the limitation inline. I think the batch-reproducibility gap deserves its own issue against the transport kernel rather than blocking this scheduling seam — happy to open one, or to dig further here if you'd prefer. What's your call?

---

tests/unit/test_osh_simulation.c
```diff
@@ -448,6 +449,278 @@ static void test_no_run_control_runs_to_completion(void) {
...
+ *   energy.  The RNG stream of every history is a pure function of its global
+ *   index (all but one neutron-producing primary are bit-identical across
+ *   batchings), so this is not a seeding defect and not neutron-pool overflow
+ *   (zero drops in both).  Rather, one primary's ion history reaches its nuclear
+ *   vertex at a fractionally different point depending on whether the run is
+ *   transported as one pass or split into batches, which flips a single stochastic
+ *   reaction and cascades.  That is a transport-kernel batch-reproducibility
+ *   limitation tracked separately from this scheduling seam; asserting exact
+ *   final-vs-LIVE equivalence here would encode a guarantee the kernel does not
+ *   yet provide.
```
grzanka, line 639: **This limitation is mis-diagnosed. It is not a floating-point kernel limitation — it's a silent ion-pool overflow. I reproduced it and traced the whole chain; the good news is the scheduling seam is fine and the RNG streams really are per-history.**

### The one-line root cause
The ion pool is sized `min(default_capacity, nstat)` (`simulation_ion_pool_capacity`, `src/simulation/osh_simulation.c:80`). For this fixture that is `min(4096, 200) = 200`. In **final-only** mode all 200 primaries are injected in one shot, so the pool is filled **to capacity with primaries** and has **zero room** for nuclear secondaries. Every recoil proton / fragment born while the pool is saturated is **silently dropped** at the ion-push guard (`src/transport/osh_transport_ion_step.c`, in `osh_transport_ion_step`):

```c
/* -- ion push -- */
if (pool->n >= pool->capacity) {
    continue;                     /* dropped: no counter, and osh_rng_split() below never runs */
}
...
osh_rng_split(&pool->rng[s], rng); /* seeds child by DRAWING 2×u64 from the PARENT stream */
```

In **LIVE K=64**, each batch injects ≤64 primaries into that same 200-slot pool, so there is always headroom and **nothing is dropped**. The number of nuclear secondaries actually transported is therefore a function of the batch size — and *that* is the entire 2.6 %.

### Evidence (nstat=200, seed 89736501, this exact fixture)
| run | ion-secondary drops | Σ(energy/nstat) |
|---|---|---|
| final-only, pool clamped to 200 | **54** | 138.2098 |
| LIVE K=64, pool clamped to 200 | 0 | 141.8082 |
| final-only, **clamp bypassed** (cap 4096) | **0** | 141.8082 |
| LIVE K=64, clamp bypassed | 0 | 141.8082 |

Give final-only the same headroom and it matches LIVE **to 1.4e-13** (141.80816020847766 vs …757) — i.e. once nothing is dropped, the two agree to pure summation-reassociation noise, *exactly* the guarantee this PR claims. Profiling confirms the same events fire in both (`nuclear_events=47`, `secondaries=144` **created** identically) — final-only just never transports 54 of them. So **LIVE is the correct result and final-only is the buggy one**, the reverse of what this comment assumes.

### Correcting each claim, in order
- *"not a seeding defect"* — ✅ correct. Per-history seeding is a pure function of the index; I verified the beam fill, `osh_rng_seed_history`, `osh_rng_split`, and pool compaction all carry the stream correctly. This is why 00_minimal (NUCRE off) is bit-identical across batchings.
- *"not neutron-pool overflow (zero drops in both)"* — ✅ true for **neutrons**, but that's the wrong pool. The overflow is in the **ion** pool, whose drop path is a bare `continue` with **no counter** (contrast the neutron path's `np->n_dropped++`). That missing counter is *precisely why* it stayed invisible and got ruled out. It **is** an overflow.
- *"one primary's ion history reaches its nuclear vertex at a fractionally different point … a transport-kernel batch-reproducibility limitation"* — real symptom, wrong cause, and second-order. Mechanism, traced: `osh_rng_split(child, parent)` seeds the child by consuming **two draws from the parent's own stream**, and only when the child is injected. A **dropped** secondary skips the split, so the parent's stream is **not advanced**, its own later draws shift, and its track diverges. This bites the primaries that **survive** a pp-elastic scatter (proton stays, recoil is the dropped secondary). Trace of primary 105 — first divergence at step 153, inputs bit-identical, only the RNG state moved:
  ```
  final  z=11.62858148468411 e=50.916896595416347 rng=0db6fc29afdcfd2b
  live   z=11.62858148468411 e=50.916896595416347 rng=29362106679d8497
  ```
  The prior step lost **18.8 MeV in one step** (69.68→50.92) — the pp-elastic recoil, dropped in final-only, injected in LIVE. The z≈13.16 vertex you cited is a *downstream* reaction flipped by this desync (the "cascade"). It's not one primary but **three** (84, 105, 195) that desync this way; 25 primaries drop a secondary overall, but the other 22 dropped from primary-*killing* channels (absorb/fragmentation), so only their deposited energy differs, not a surviving track.

### Bottom line
Two defects, one root cause — see my note on the reproducibility assertion below for the fix directions. The kernel and the RNG model are not at fault; I'd correct this comment rather than ship it, since as written it enshrines "kernel FP limitation" as the explanation.

---

tests/unit/test_osh_simulation.c
```diff
@@ -448,6 +449,278 @@ static void test_no_run_control_runs_to_completion(void) {
...
+    ASSERT_TRUE(completed_live_a == 200ull);
+    ASSERT_TRUE(energy_live_a > 0.0);
+    ASSERT_TRUE(prof_live.neutrons_banked > 0ull);
+
+    /* Reproducibility of a fixed cadence: the same LIVE policy re-run reproduces
+     * the scored energy exactly (bit-for-bit), the determinism guarantee the
+     * count cadence is meant to provide. */
+    run_checkpoint_case("06_minimal_nucre", 200ull, 64ull, &completed_live_b, &energy_live_b, NULL);
+    ASSERT_TRUE(completed_live_b == 200ull);
+    ASSERT_TRUE(energy_live_b == energy_live_a);
```
grzanka, line 672: Worth being clear about what this assertion buys you: `energy_live_b == energy_live_a` only proves that a **fixed** cadence is deterministic — which it trivially is, because for a fixed K the drop pattern is identical on every run. It structurally *cannot* catch the real bug, which is that the drop pattern (and therefore the answer) **changes with K**. A sharper regression is the one that currently "does not hold": assert final-vs-LIVE agreement **after** giving the ion pool headroom for secondaries — which does hold, to 1.4e-13.

### The two defects (one root cause) and how I'd fix them

**1. Sizing (dominant, the −2.6 %).** `simulation_ion_pool_capacity()` returns `min(cap, nstat)`, so a full-nstat wavefront leaves no room for the secondaries that same wavefront produces. Options, roughly in order of preference:
   - Size the ion pool with explicit headroom for in-flight secondaries (e.g. `nstat + expected_secondary_slack`, or just don't clamp the default 4096 down to nstat — the clamp is what removes the headroom here).
   - At minimum, **add an ion-drop counter** (mirror `np->n_dropped`) and surface it, so a silent physics loss can never again masquerade as a "kernel FP limitation." Right now a dropped recoil proton is completely unobservable.

**2. Reproducibility (why it presents as "the vertex moved").** `osh_rng_split(child, parent)` advances the **parent's** stream, and only at inject time — so a drop, or any reordering/overflow, silently perturbs the parent. The drop-independent design is already written down in #148: seed child streams from `(parent prim_idx, secondary ordinal)` — `osh_rng_split(child, parent_seed, parent_idx, ordinal)` — instead of by drawing from the parent. Then a dropped/reordered/overflowed secondary can never move the parent's sequence, and reproducibility stops depending on pool occupancy at all. That's the property that makes the result truly batch-invariant regardless of capacity.

Fix #1 makes this fixture correct today; fix #2 makes it *robust* (so the same class of bug can't reappear the first time a large-nstat run legitimately overflows the pool). Happy to open both as issues against the transport kernel and wire the sharper final-vs-LIVE assertion in here once #1 lands.

---

grzanka (commented):
Dug into the ~2.6 % final-vs-LIVE discrepancy this PR documents as a "transport-kernel batch-reproducibility limitation." **It isn't one.** I reproduced it (nstat=200, this fixture: final-only 138.2098 vs LIVE K=64 141.8082) and traced it to a **silent ion-pool overflow** — not the kernel, and not the RNG streams, which are correctly per-history and order-independent.

Root cause in one line: the ion pool is sized `min(cap, nstat)`, so final-only fills it entirely with primaries and drops **every** nuclear secondary born while it's saturated (54 here, via a bare `continue` with no counter); LIVE K=64 leaves headroom and drops none. Give final-only the same headroom and it matches LIVE **to 1.4e-13**. So **LIVE is the correct answer and final-only is the buggy one** — and the author's cited "one primary's vertex moved" is a real but second-order symptom: `osh_rng_split()` consumes the *parent's* draws only when the child is injected, so a dropped recoil desyncs the surviving parent's stream.

Full mechanism, the measured evidence table, the primary-105 divergence trace, and the two fix directions are in the two inline comments. Net recommendation: the scheduling seam here is good — I'd **correct the "documented limitation" comment** (it enshrines the wrong cause) and split the two real defects (pool sizing + drop-independent child seeding, per #148) into their own issues rather than blocking this PR. I have not touched any code — reporting first, as asked.

---

tests/unit/test_osh_simulation.c
nbassler: Different suggestion: keep the secondary ions in a dedicated pool allocated for them. Then the primary pool would never overfill. They also behave most likely quite different than the primary pool, since the species and energies are more mixed, increasing risk of cache misses?

Also notice: primary particles can be neutrons if user simulates neutron beams or PHSP files, which is not unlikely.

Dropping particles when overfilling is not a good option. Then rather stop, and realloc(). With some careful default  tuning, reallocs can be made very rarely needed.

Notice nuclear fragmentation / spallation reactions may cause many more secondaries than primaries in a single reaction. (We do not need to be able to treat fission, that would require a completely different approach anyway due to the exponential increase of secondaries.)

grzanka: Ok, but may be in new PR?

nbassler: sure!

PR #208, opened 2026-07-01
190 physics add vavilov and landau straggling

### LINKED ISSUE #190: PHYSICS: add Vavilov and Landau straggling
- reduce energy cutoff to 25 keV/nucl
- establish reference for comparisons for all STAGL modes
- investiage peak width but also dig into LET scoring artefacts at Bragg peak.

nbassler:
see #190


nbassler:
current situation covered here in report 
[straggling_report.pdf](https://github.com/user-attachments/files/29550837/straggling_report.pdf)


## STRAGL 0 (off):

image

small range difference, also noticed in PR #133. This is due to better interpolation dE/dx tables in OSH. 

## STRAGL 1 (Gaussian)
image


Is OK in terms of width, dip in DLET and TLET before last particle stops is single particle noise (compare to fluence and dose curves, the noise happens far after the BP)

grzanka:
## Detailed review of #208

> **Disclosure:** This review was produced **intentionally / on purpose** by **Claude Opus 4.8** (Anthropic's Claude Code), acting as an automated reviewer at the request of the repository owner. Findings were verified against the code on the PR branch and by independently recomputing the physics; please still apply human judgment before acting on them.

### What this PR actually is

Despite the title *"add vavilov and landau straggling"*, **no Vavilov or Landau sampler is added here.** This is a clean, well-scoped **infrastructure / preparation** PR for #190:

- Splits the straggling module into a **dispatcher + per-model** layout mirroring `scat` (`osh_physics_straggling{,_gauss}.{c,h}`), moving the Bohr Gaussian width into `_gauss` **byte-for-byte**.
- Adds shared straggling **kinematics helpers**: ξ, `E_max`, κ = ξ/`E_max`, and the reduced mean λ̄.
- **Reserves STRAGG 2 (Vavilov) and STRAGG 3 (Urbán)** — parsed, mapped through, and *rejected* at transport setup with `OSH_ENOTSUP`.
- Lowers the hard transport cutoff to **25 keV/u** (aligned to the LOADDEDX table lower bound).
- Adds **STRAGG 0/1/2 benchmark decks** (+ SH12A mirrors) and a `plot_straggling.py` overlay tool.

The commit messages and READMEs are honest that the samplers land later; the TODO.md breakdown is excellent. Overall this is **high-quality scaffolding — recommend merge after the title/scope and the small nits below.**

---

### ✅ Verified correct

I independently recomputed the new kinematics against the PDG "Passage of particles through matter" definitions (proton, T=200 MeV, water) — all four helper values match the unit-test references:

| quantity | formula | computed | test asserts |
|---|---|---|---|
| `E_max` | 2mₑc²β²γ² / (1 + 2γmₑ/M + (mₑ/M)²) | 0.481493 | 0.4815 ✓ |
| ξ | (K/2)(Z/A)(z²/β²)·d | 0.026589 | 0.026590 ✓ |
| κ | ξ / `E_max` | 0.055223 | 0.055225 ✓ |
| λ̄ | −β² − ln κ + (γ_E − 1) | 2.15306 | 2.1529 ✓ |

- **λ̄ is a good choice**: −β² − ln κ + (γ_E − 1) is the analytic mean of the Vavilov λ variable, so `ΔE = ξ·(λ − λ̄)` is mean-preserving by construction — the future sampler won't double-count the CSDA mean. Nicely documented.
- **STRAGG 0/1 byte-identity holds.** `osh_physics_straggling_gauss.c` keeps the exact `BOHR_SQRT_C 0.396128` literal and identical input guards. The literal is intentionally *not* recomputed from the new (more precise) `const.h` electron mass — true √(K·mₑ) ≈ 0.396125 — and the comment explains why. Good discipline; the "unchanged on the 200 MeV benchmark" claim is credible.
- **STRAGG 2/3 wiring is consistent across every layer**: `beam_defs.h` (macro + name), `osh_transport.h` / `osh_physics_straggling.h` enums, `osh_beam_parse.c` (range 0–3), `osh_simulation.c` (map), `osh_transport_ion.c` (reject). `osh_beam_stragg_names[]` was correctly extended to 4 entries, so `osh_beam_stragg_names[straggl]` stays in bounds.
- Unit tests are picked up automatically (`file(GLOB test_*.c)`) and run via the no-arg `main()`. CI is green on all platforms.

---

### Findings (most to least important)

**1. Title/description overstate the scope.** (Copilot flagged this too — I concur.) Nothing here samples Vavilov or Landau. Suggest retitling to something like *"straggling: split module + shared κ/ξ/E_max/λ̄ kinematics; reserve STRAGG 2/3 (prep for #190)"* so the history reflects the actual delta.

**2. Hard-coded `0.025` cutoff duplicates `OSH_MATERIAL_RUNTIME_EMIN`.** (Copilot, `osh_transport_ion_step.c:52`.) I checked this is safe: `OSH_MATERIAL_RUNTIME_EMIN` is defined `0.025` **in the same MeV/nucleon units**, the file already includes `material/runtime/osh_material_runtime.h`, and the cutoff is applied as `cutoff_total = OSH_TRANSPORT_ION_EMIN_MEV_PER_U * a_proj` (`:982`) — per-nucleon throughout. Since the comment itself says the floor *should equal* the table lower bound, referencing the macro (keeping the `_MEV_PER_U` name to signal units) prevents future drift. Minor, but worth it.

**3. Stale wording in the `OSH_BEAM_TMIN` rejection message.** `osh_beam_parse.c:~1029` still rejects primary-beam energies below `OSH_BEAM_TMIN` (0.1) with *"is below transport threshold %.4f MeV/nucleon"*. The transport threshold was just decoupled and lowered to 0.025, so the message now misstates it. Keeping a 0.1 MeV/u floor on the *primary beam* is defensible, but the text should stop calling 0.1 the "transport threshold" (e.g. "below minimum beam energy").

**4. `TODO.md` commits a machine-local path.** `(plan: ~/.claude/plans/ethereal-enchanting-petal.md)` won't resolve for anyone else. Drop it or replace with the #190 link.

**5. `osh_physics_straggling_lambda_bar` uses `0.0` as its invalid-input sentinel**, but `0.0` is also a legitimate λ̄. Harmless today (no caller consumes it yet), but fragile once the sampler lands — consider `NAN` or an explicit out-param.

**6. Minor design notes for the follow-up sampler:**
   - `enum osh_straggling_model` (new) manually mirrors `enum osh_transport_straggling_mode` and is **currently unused**. A `static_assert` cross-checking the two, or a pointer in each comment, would catch silent drift when the dispatcher is wired.
   - `osh_physics_straggling_xi()` takes β² as an argument while `osh_physics_straggling_emax()` recomputes β² internally from (T, M). When combined into κ, an inconsistently-computed β² from the caller would make κ marginally inconsistent. Consider deriving β² from one source.

---

### Status / coverage (not defects, just so reviewers know)

- **The three new IDD benchmark cases don't gate anything yet.** None ships a `reference/` curve, and `tests/reference/CMakeLists.txt` reports curve-less cases as **SKIPPED** ("infrastructure can precede the data"). So the ~42 new benchmark files add **no active regression check today**; `strag2` will additionally **FAIL-by-design** (`ENOTSUP`) once a curve lands — the intended red-test driver for #190. All clearly documented in the READMEs.
- **codecov** flags the STRAGG 3 stub lines in `osh_simulation.c` and `osh_transport_ion.c` as 0% covered. A tiny test feeding `STRAGG 3` and asserting `OSH_ENOTSUP` (and `STRAGG 9` → parse error) would lock in the "reserved-but-rejected" contract and close the gap.

### `tools/plot_straggling.py`

Dev/analysis utility (not shipped, not on the CI test path). Well-structured: graceful handling of an unavailable code, per-case timing, robust numeric loader that skips `# ! *` comment lines and stray label rows. No blocking issues. Nit: `set_nstat()` rewrites the `NSTAT` line to a fixed `"NSTAT {n} -1"`, silently dropping any original second field — fine for this tool's purpose, just noting.

---

**Verdict:** Correct, careful, well-documented infrastructure. Nothing here is a blocker. Addressing #1 (title) and #2–#4 (small consistency/hygiene nits) would make it a clean merge; #5–#6 are worth keeping in mind for the sampler PR.

---

nbassler:
[straggling_report.pdf](https://github.com/user-attachments/files/29573993/straggling_report.pdf)
image

nbassler:
image

nbassler:
image

image
[straggling_report.pdf](https://github.com/user-attachments/files/29574899/straggling_report.pdf)

nbassler:
## Cross-validation: OSH Vavilov sampler vs Thomsen's original `libvav`

Compared the OpenShieldHIT clean-room Vavilov evaluator
(`osh_physics_strag_vavilov_lambda`) directly against Bjarne Thomsen's original
Fortran `vavinv` (`_temp_shieldhit/libvav_compare/`), over a
21 κ × 6 β² × 20 u grid (2490 points, skipping the region outside libvav's
stated validity), and adjudicated the disagreements against the **exact** Vavilov
distribution. (Both codes cover the same κ ∈ [0.01, 10] band here and both hand
off to a Landau sampler — `laninv` in SH12A, our `_landau` in OSH — for κ < 0.01,
so that regime is out of scope for this comparison.)

### Overall agreement — excellent
| metric | OSH − libvav |
|--|--|
| mean | −0.011 |
| rms | 0.047 |
| max \|Δλ\| | 0.339 |

By κ-band the rms is **~0.012–0.022 for κ ≥ 0.11** (a tie), degrading only in the
smallest band (κ < 0.04: rms 0.085). Disagreements are concentrated entirely at
the **corners**: smallest κ (0.01–0.04) and extreme u (→0 or →1).

### Adjudication vs the exact distribution — Thomsen is more accurate at the corners
At the worst points, libvav is within ~0.01–0.03 of the exact Vavilov; OSH is off
by 0.2–0.3. (At κ = 0.01 the distribution is essentially Landau, whose inverse CDF
is known exactly — libvav matches it to ~0.004.)

| κ | β² | u | OSH | libvav | **exact** |
|--|--|--|--|--|--|
| 0.01 | 0.50 | 0.001 | −2.964 | −2.625 | **−2.644** |
| 0.01 | 0.05 | 0.003 | −2.723 | −2.407 | **−2.417** |
| 0.08 | 0.05 | 0.995 | 16.494 | 16.196 | **16.170** |
| 0.05 | 0.05 | 0.995 | 21.642 | 21.882 | **21.855** |

### Root cause — a fixable reference-generation artifact, not the fit/runtime
A freshly computed exact value at high resolution gives −2.644 (matching libvav),
but OSH produces −2.964. OSH was fit to our exact-Vavilov reference cache, which
had **glitchy values at the small-κ / extreme-u corners** (the numerical-roundoff
region of the quadrature). The fit faithfully reproduced them. Fix: robustify
`tools/vavilov_fit/vavilov_exact.py` near λ_min for κ→0.01, regenerate the cache,
re-fit — expected to collapse the corner error from ~0.3 to ~0.02.

### Verdict
Agreement is excellent everywhere real transport operates (κ ≳ 0.1 → rms ~0.02).
Thomsen's fit is genuinely more accurate at the extreme small-κ corners, but that
is a fixable artifact in OSH's reference generation, and those corners are never
reached in practice (measured κ_min for 200 MeV p in water is ~0.12; ξ is tiny at
small κ, so the 0.3 λ gap is a sub-keV effect on a sliver of the thinnest steps).
OSH's differentiator over `libvav` is provenance — its own coefficients fit to the
published exact distribution — not accuracy.

nbassler:
## Update: small-κ reference fix, evaluator speedup, review comments addressed

Follow-up since the last review. Four commits (`630522f`, `b2bd15b`, `2e1561c`, `0e4cba8`).

### What changed

- **Fixed the small-κ Vavilov reference glitch at its source.** The exact-Vavilov
  generator (`tools/vavilov_fit/vavilov_exact.py`) scaled the λ-grid's low-loss
  edge by `6·sd`, which diverges as κ→0 (edge ≈ −50 at κ=0.01). A uniform
  400-point grid then wasted almost all its points in a dead zone and
  under-resolved the peak, so the inverse CDF at extreme-low `u` read ≈ −3.0
  instead of the true ≈ −2.64. Replaced with a tight, κ-independent left edge +
  dense linear core + geometric heavy-tail sampling + a relative pdf noise floor.
  Regenerated the cache/fixtures at a finer `n=6000` mesh (generation now
  parallelised).
- **Sped up the hot-path evaluator** (`osh_physics_strag_vavilov_lambda`) — see
  below.
- **Tightened** `VAV_TOL` 0.4 → 0.25 with documented rationale.
- **Addressed the outstanding Copilot review comments** (cutoff-macro
  duplication, `lambda_bar` NaN sentinel, `vavilov_exact.py` y=0 integrand guard,
  runtime domain-clamp + honest header contract, and the stale
  "Vavilov not implemented" docs). Threads resolved.

### Accuracy vs the exact Vavilov distribution (OSH vs SH12A / Thomsen `libvav`)

The low-κ / extreme-low-`u` corner is now essentially exact, and the only
residual is confined to the extreme high-loss tail — where **OSH is an order of
magnitude closer to the exact distribution than the reference `libvav`**:

Fit error vs exact, by `u`-band (all κ, β²):

| region | worst \|Δλ\| |
|---|---|
| κ ≳ 0.1 (where transport operates) | < 1e-3 |
| u ≤ 0.95 | 0.016 |
| u ≤ 0.99 | 0.073 |
| u ≤ 0.995 (incl. extreme tail) | 0.202 |

Direct OSH vs `libvav` vs exact at the worst (small-κ) tail points
(`libvav` built with gfortran and evaluated on the same grid):

| κ | β² | u | exact λ | OSH \|Δ\| | libvav \|Δ\| |
|---|---|---|---|---|---|
| 0.063 | 0.05 | 0.995 | 18.78 | **0.20** | 5.28 |
| 0.040 | 0.30 | 0.995 | 24.43 | **0.15** | 6.09 |
| 0.010 | 0.80 | 0.995 | 51.60 | **0.12** | 13.47 |
| 0.010 | 0.99 | 0.995 | 46.32 | **0.09** | 18.75 |
| 0.063 | 0.05 | 0.950 | 10.63 | **0.02** | 3.64 |
| 0.251 | 0.05 | 0.950 (sanity, normal region) | 4.64 | 0.001 | 0.001 |

`libvav`'s own stated validity is `u ≤ 0.990` for κ < 0.04, so those tail points
are outside its range; in the normal region both agree to ~1e-3. The residual is
a fit-form limit, and it lands on a rare-κ × 0.5%-tail region (sub-keV; edge of
the sampler range) — never reached in practice (measured κ_min ≈ 0.12 for 200 MeV
p in water).

### Runtime speed

Micro-benchmark, 5×10⁷ calls, `-O2`, identical RNG baseline subtracted:

| sampler | net ns/call |
|---|---|
| OSH — before (double, 3× `log`) | ~76 |
| **OSH — now (double, precomputed band-log constants + `logf` transforms)** | **~60** |
| `libvav` `vavinv` (`REAL*4`) | ~51 |

**~20% faster, staying fully double precision, output physically unchanged.** The
win was *not* precision — a faithful `float` port of the evaluator measured
~75 ns (no gain; scalar `float`/`double` arithmetic costs the same, and float
would only spend accuracy margin: max |float − double| ≈ 3e-5 λ vs the 0.2 fit
error). The real cost was two redundant `log()` calls per invocation on per-band
**constants**:

| variant (all double) | net ns | Δ vs exact evaluator |
|---|---|---|
| current (3× `log`) | 78.0 | — |
| + precomputed `ln(klo)`, `ln(khi)`, `1/ln-span` | 65.5 | bit-identical (3.6e-15) |
| + `logf`/`log1pf` transforms | 62.1 | 2.75e-5 λ |

The remaining gap to `libvav` is purely its `REAL*4` arithmetic, which buys
nothing here on accuracy or scalar speed.

<sub>Analysis and patch assisted by Claude Code.</sub>

nbassler:
image

PR #164, opened 2026-06-19
docs: one command per copy-pasteable code block

grzanka:
## Summary

Fixes #155.

The rendered documentation (and GitHub markdown) attaches a one-click **copy** button to every fenced code block. When a single block bundled several *independent* or *alternative* commands, that button copied all of them at once — so a user who wanted just one command was forced back to selecting text with the mouse, defeating the purpose of the button.

This change splits those blocks so that each copyable block holds **a single command**, each introduced by a short prose lead-in describing what it does. Genuine run-this-in-order sequences (e.g. `git clone` → `cd` → configure → build) are deliberately kept together, since copying them as one unit is the intended workflow.

The result matches the style already used in `docs/user/command-line.md` and `PACKAGING.md`, so the docs are now consistent throughout.

## Files touched

| File | What changed |
|------|--------------|
| `docs/user/getting-started.md` | Split the *Install system-wide* (3 alternatives) and *Run a case* variations (4 alternatives) blocks |
| `docs/index.md` | Split *Quick start* — build sequence kept together, run variations separated |
| `docs/dev/build.md` | Split the three alternative build configurations (release / debug / coverage) |
| `docs/dev/testing.md` | Split the three alternative `ctest` invocations |
| `README.md` | Split *How to build*, *Install*, and the application-invocation block |
| `DEVELOPER.md` | Split the four-command preset block into per-command blocks |

No content was removed — every command and its explanation is preserved, just reorganised. `command-line.md`, `PACKAGING.md`, and `CONTRIBUTING.md` already followed this style and were left unchanged.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---
_Generated by [Claude Code](https://claude.ai/code/session_01JYQfov3Nrr6NDZqGSkg15x)_

PR #210, opened 2026-07-01
docs: add CLAUDE.md agent brief and thin agent-config pointers (#205)

### LINKED ISSUE #205: Add CLAUDE.md file
And any agentic configs as well

grzanka:
Closes #205.

## Why

Issue #205 asks for a `CLAUDE.md` (and "any agentic configs as well"). The
discussion on the issue already set the direction:

- **@nbassler** was wary this could "get very messy" and would rather lean on the
  existing `llms.txt`.
- **@grzanka** laid out the distinction: `llms.txt` tells *any* AI *what the
  project is and where to look*; a `CLAUDE.md` tells an agent *how to behave while
  modifying the code* — and, crucially, since we've already invested in
  `DEVELOPER.md`, it "would probably just point Claude at that file rather than
  repeating it."

This PR follows that steer literally. It's the same "one source of truth, thin
signpost, don't duplicate" reasoning @grzanka argued for the `INSTALL.md`
question on #164 — applied to agent config.

## What

- **`CLAUDE.md`** — a lean, one-screen operating brief for coding agents. It is a
  *signpost*, not a second copy: a table routes to `DEVELOPER.md` (style
  authority), `llms.txt` (project map), `CONTRIBUTING.md`, `docs/`, and per-module
  READMEs. It then captures only the high-value, repo-specific things an agent
  gets wrong:
  - verified build / test / format commands, one command per copy-block (the
    convention from #164);
  - the handful of hard rules agents break most (K&R block-top declarations, no
    `typedef struct`, no `//` in production, `double const *p`, banned POSIX-only
    APIs, no hot-path allocation, predicate-`int` vs `enum osh_status`), each
    pointing back to `DEVELOPER.md` for the full set;
  - module layout and how the auto-discovered `tests/unit/test_*.c` harness works;
  - the SH12A zero-shared-code provenance note.
- **`AGENTS.md`** and **`.github/copilot-instructions.md`** — thin pointers to
  `CLAUDE.md` so other tools (and GitHub Copilot, which is actively used on this
  repo for reviews and `@copilot` tasks) land on the same page. They are
  redirects, not copies, so nothing can drift. Copilot's file additionally inlines
  the few C rules it most needs at generation time.

Addresses the "messy" concern by keeping exactly **one** place with real content;
the other two files are 2–15 line redirects.

## Accuracy

Commands were checked against `CMakePresets.json` and the CI workflow, not copied
from prose. One correction worth flagging: there are **no CTest presets**, so
`ctest --preset debug` errors with *"No such test preset"* — `CLAUDE.md` documents
the working form `ctest --test-dir build_debug` instead. (Copilot flagged this same
issue on #164; `llms.txt` still carries the outdated `ctest --preset debug` and a
stale `build_rel/` release path — happy to fix those in a small follow-up if
wanted, kept out of scope here.)

## Notes

Docs-only; no code paths touched, so CI is build/format/tidy + the existing test
suite.

https://claude.ai/code/session_013sX2yXJShQJn8HV3iergoQ

---
_Generated by [Claude Code](https://claude.ai/code/session_013sX2yXJShQJn8HV3iergoQ)_

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -1,0 +1,17 @@
+# GitHub Copilot instructions
+
+The agent brief for this repository lives in **[`../CLAUDE.md`](../CLAUDE.md)**.
+Follow it (and the docs it links — [`DEVELOPER.md`](../DEVELOPER.md),
+[`llms.txt`](../llms.txt)) when suggesting or reviewing code here.
+
+The rules Copilot most often needs when generating C for this project:
+
+- C11, but declare variables at the **top of each block** (K&R); none inside `for`/`if`.
+- Use `struct foo` explicitly — **never** `typedef struct`.
+- Block comments `/* */` only; `//` is reserved for temporary notes.
+- `double const *p`, not `const double *p`.
+- No POSIX-only APIs (`strcasecmp`, `mkdtemp`, `getpid`, `<unistd.h>`, `<threads.h>`, …) — Windows is a target.
+- No heap allocation on the simulation hot path.
+- Predicates return `int` (1 = yes); operations return `enum osh_status` (`OSH_OK = 0`).
+
+See [`../CLAUDE.md`](../CLAUDE.md) for the full list and the build/test/format commands.
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -4,14 +4,15 @@ The agent brief for this repository lives in **[`../CLAUDE.md`](../CLAUDE.md)**.
 Follow it (and the docs it links — [`DEVELOPER.md`](../DEVELOPER.md),
 [`llms.txt`](../llms.txt)) when suggesting or reviewing code here.
 
-The rules Copilot most often needs when generating C for this project:
+The rules Copilot most often needs when generating C for this project (each tagged
+with its authoritative section in [`DEVELOPER.md`](../DEVELOPER.md) — that § wins):
 
-- C11, but declare variables at the **top of each block** (K&R); none inside `for`/`if`.
-- Use `struct foo` explicitly — **never** `typedef struct`.
-- Block comments `/* */` only; `//` is reserved for temporary notes.
-- `double const *p`, not `const double *p`.
-- No non-portable APIs (POSIX-only or missing on MSVC): avoid `strcasecmp`/`<strings.h>`, `mkdtemp`, `getpid`, `<unistd.h>`, `<threads.h>`, … — Windows is a target.
-- No heap allocation on the simulation hot path.
-- Predicates return `int` (1 = yes); operations return `enum osh_status` (`OSH_OK = 0`).
+- C11, but declare variables at the **top of each block** (K&R); none inside `for`/`if`. — §1
+- Use `struct foo` explicitly — **never** `typedef struct`. — §2.1
+- Block comments `/* */` only; `//` is reserved for temporary notes. — §4.1
+- `double const *p`, not `const double *p`. — §2.2
+- No non-portable APIs (POSIX-only or missing on MSVC): avoid `strcasecmp`/`<strings.h>`, `mkdtemp`, `getpid`, `<unistd.h>`, `<threads.h>`, … — Windows is a target. — §11.1
+- No heap allocation on the simulation hot path. — §10
+- Predicates return `int` (1 = yes); operations return `enum osh_status` (`OSH_OK = 0`). — §5
 
 See [`../CLAUDE.md`](../CLAUDE.md) for the full list and the build/test/format commands.
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -10,7 +10,7 @@ The rules Copilot most often needs when generating C for this project:
 - Use `struct foo` explicitly — **never** `typedef struct`.
 - Block comments `/* */` only; `//` is reserved for temporary notes.
 - `double const *p`, not `const double *p`.
-- No POSIX-only APIs (`strcasecmp`, `mkdtemp`, `getpid`, `<unistd.h>`, `<threads.h>`, …) — Windows is a target.
+- No non-portable APIs (POSIX-only or missing on MSVC): avoid `strcasecmp`/`<strings.h>`, `mkdtemp`, `getpid`, `<unistd.h>`, `<threads.h>`, … — Windows is a target.
 - No heap allocation on the simulation hot path.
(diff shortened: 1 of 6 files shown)


nbassler:
By "messy" I meant more files cluttering the project root. This #PR should then also allow `.codex`, ... and whatever LLM we will use in future. I had hoped that llms.txt would be the catch all.

grzanka:
Accepting more files (I know it may look like a mess) may improve the vibe-coding. llms.txt is more for chat-like LLMs which crawl the web to learn about something. So for vibe-coding you need different stuff (which may have links to other files) than for general purpose query of future users who may like to ask in chat windows in web browser: "explain to me what openshieldhit does"

nbassler (changes requested):
My concern is not only that this adds more root/config files. The bigger problem is that it creates several partly duplicated summaries of the same rules:

- `DEVELOPER.md` has the full coding rules.
- `llms.txt` already repeats a subset of them.
- This PR adds `CLAUDE.md` with another selected subset.
- `.github/copilot-instructions.md` then repeats a smaller subset again.

That is hard to maintain. We can already see the failure mode: `llms.txt` has stale build/test guidance (`ctest --preset debug`, binary-dir notes), while this PR fixes the same topic only in `CLAUDE.md`. This is exactly the kind of drift I am worried about.

I am not arguing that `DEVELOPER.md` should become short or terse. Quite the opposite: I think `DEVELOPER.md` should remain the complete, explanatory, pedagogical source of truth. Many contributors, including students, need to understand not only what the rule is, but also why the project made that design choice.

What I would prefer is to first consolidate `DEVELOPER.md` into a clearer numbered ruleset. The rules can be grouped logically, and each rule can still contain prose explanation, motivation, examples, and exceptions where needed. For example:

### §1 C dialect, declarations, and initialization

Rule: OpenShieldHIT uses C11, but generally keeps declarations at the beginning of the block. Do not declare variables inside `for`, `if`, or other control statements.

Reason: This K&R-inspired style makes review easier because the local state of a function is visible at the start of each block. We do not choose this because modern C cannot support mixed declarations; we choose it because the project values a consistent, easily scanned style.

Exceptions: Some values must be initialized at declaration for correctness. For example, `const` locals must receive their value when declared, and some API-owned objects may need defensive initialization before being passed to cleanup or overwrite functions. These exceptions should be used when they improve correctness or clarity, not as a way to reintroduce mixed declaration style.

That kind of structure would give us one place to maintain the real rules, while still allowing `DEVELOPER.md` to explain the motivation in enough detail for humans to learn from it. It also makes reviews easier: we can say “this violates `DEVELOPER.md §1`” instead of quoting whichever duplicated prose an agent happened to read.

Then tool-specific files can be thin entrypoints only:

- `AGENTS.md`: “Coding rules live in `DEVELOPER.md`; project overview in `llms.txt`.”
- `CLAUDE.md` and any `.codex` guidance: same, maybe with workflow-specific pointers, but no duplicated rule bodies.
- `.github/copilot-instructions.md`: same.

So my suggestion is: do not add another rule summary here. First make `DEVELOPER.md` the numbered, explanatory source of truth, then add only minimal redirects from agent-specific files if we decide those files are worth having.

nbassler:
i will make a new developer.md (working on it)

nbassler:
One more point in favor of the numbered `DEVELOPER.md` approach: it should also be easier for LLM-based tools to parse and follow.

Why this helps:

- **Stable anchors:** references such as `DEVELOPER.md §7.2` are much easier to retrieve, cite, and obey than an unstructured prose section.
- **Clear instruction hierarchy:** separating `Rule`, `Reason`, `Exceptions`, and examples helps distinguish normative instructions from explanation.
- **Less duplication:** agent-specific files can point to precise sections instead of carrying their own shortened copies.
- **Lower drift risk:** if a rule changes, there is one authoritative place to update it.

So this is not about making `DEVELOPER.md` sparse. It can stay explanatory and pedagogical, but with a structure that is easier for both humans and tools to navigate.

nbassler:
I think we should distinguish tool configuration from project guidance.

- `.claude/` and `.codex` are tool/runtime configuration areas. They can contain permissions, local settings, hooks, or machine-specific behavior. These should usually not be project policy documents, and in this repo they are currently ignored.
- `CLAUDE.md`, `CODEX.md`, and `AGENTS.md` are Markdown guidance files. If we add them, they should be thin pointers to the authoritative docs, especially `DEVELOPER.md`, not duplicate rule summaries.

So I would avoid adding or encouraging multiple tool-specific Markdown rule copies. If a tool needs an entrypoint, it should say roughly: “Project rules live in `DEVELOPER.md`; project overview lives in `llms.txt`; contribution workflow lives in `CONTRIBUTING.md`.”

That keeps config separate from policy and avoids needing `.claude/`, `.codex`, `CLAUDE.md`, `CODEX.md`, `AGENTS.md`, and Copilot instructions to all restate overlapping rules.

so we should take a closer look at `.gitignore` too.

nbassler:
llms.txt needs to be fixed as well then

grzanka:
I like the direction of your changes

nbassler:
@grzanka would you fix `CLAUDE.md`, `llms.txt`, `.gitignore` and any copilot files? Then you can merge right away. I would on my side add any codex/mistral files opportunistically in an unrelated furture PR.

grzanka:
@nbassler done — pushed `f998d70` to this branch. Fixed all four:

- **`CLAUDE.md`** and **`.github/copilot-instructions.md`** — every inline rule now cites the `DEVELOPER.md` section that owns it (`§1`, `§2.1`, `§4.1`, `§5`, `§6.1`, `§10`, `§11.1`, `§11.2`), stated as "the § is the authority". They stay useful at generation time but can't drift — each line is now a citation into your numbered ruleset, and `→ Layout` became `§9`.
- **`llms.txt`** — fixed the stale build/test guidance: `ctest --preset debug` → `ctest --test-dir build_debug` (no CTest presets exist), and the release/debug binary dirs were backwards, now `build/bin` vs `build_debug/bin` per `§14.1`. Replaced the duplicated "Key conventions" bullets with a routing table into the numbered sections, so there's no rule body left to go stale.
- **`.gitignore`** — documented and normalised the config-vs-guidance split you described: tool-config dirs (`.claude/`, `.codex/` + variants) stay ignored; the guidance Markdown (`CLAUDE.md`, `AGENTS.md`, `copilot-instructions.md`) is tracked. Verified both behaviours.

`AGENTS.md` was already a pure pointer, so left untouched. Ready to merge; I'll leave the codex/mistral files to your future PR as you suggested.

---

nbassler (approved):

PR #244, opened 2026-07-05
refactor: enhance nuclear elastic cross section calculation

### LINKED ISSUE #241: PERF: single σ_R rate loop — drop double-Tripathi on the nuclear hot path (#226 Goal 1)
Parent: #226 (SoA / SIMD / GPU-portable nuclear hot path). This is **Goal 1 — the quick win**.

## Problem
Since the #212 Stage-3 p+A elastic channel landed, the parametric Tripathi σ_R is
evaluated **twice per step** for every Z≥2 target element: once in the inelastic
rate sum and again inside the p+A-elastic rate sum (`osh_nuclear_elastic_sigma`
calls `osh_nuclear_tripathi_sigma` with identical `(zp, ap, zt, at, E/u)` and
multiplies by `SIGMA_FACTOR`). This regressed nucre1/nucre2 runtime.

## Fix
Merge the inelastic and p+A-elastic **per-element rate passes** in
`osh_nuclear_handler_step()` into a single loop that evaluates Tripathi **once**
per (element, E) and feeds both hazards:
- `rate_inel += mf / λ(A, σ_R)` when the inelastic channel is active for the element
- `rate_pa   += mf / λ_el(A, FACTOR·σ_R)` when the p+A-elastic channel is active

Evaluate Tripathi only when at least one channel needs it for that element (guard on
the skip predicates first, so hydrogen-only steps don't pay for a discarded call).
The pp-elastic (H-only) rate pass carries no Tripathi call and is untouched.

## No physics change (bit-identity guard)
- Tripathi is a pure function of its args → same inputs, same bits; the elastic pass
  already passes identical args, so reuse is exact.
- `SIGMA_FACTOR == 1.0`, `1.0·x == x` in IEEE754; keep the explicit `FACTOR·` so it
  stays correct if the macro is ever retuned.
- Per-element addition order into each accumulator is preserved → `rate_inel` and
  `rate_pa` reproduce to the bit; no new RNG draws → stream unchanged.

Out of scope for Goal 1: the post-event channel-selection loops. They re-derive σ
against a fresh RNG threshold and run only when an event fires (rare); a follow-up
could share the same cached σ, but that is a larger change with its own correctness
argument.

## Acceptance
- nucre1/nucre2 scorer output **byte-identical** before vs after.
- Record the release-build runtime delta on nucre1/nucre2 (the named regression).
- One file touched: `src/physics/nuclear/osh_nuclear_handler.c`; no API change.

Relates to #212 (regression source), #211 (hot-path perf + LTO), #219 (roofline).

nbassler:


nbassler:
Bit-identical across nucre0–3, +8.3% on nucre1, 120 tests pass, clang-format + clang-tidy clean, DEVELOPER.md §1.2 style, no in-code refs to the issue being closed.

Issue #227, opened 2026-07-03
SCORING: capture LET for point-deposited sub-threshold recoils/fragments (score_point)


nbassler:
**Follow-up to #212 (score_point origin: #179).**

## Context
Heavy recoils and p+A-elastic recoils are frequently born **below** the ion transport
threshold (0.025 MeV/nucleon). These are deposited locally via `score_point` (#179),
which currently books **energy + dose only — not LET**. So the single highest-LET
component of the field is invisible in the DLET/TLET scorers, and the recoil-LET
payoff of #212 Stages 1 (fragment transport) and 3 (p+A elastic) cannot be seen.

## Goal
Feed point-deposited sub-threshold recoils/fragments into the LET accumulators with a
representative dE/dx, so heavy-recoil / elastic-recoil LET is captured.

## Approach
- Give `score_point` the recoil's stopping power at its birth energy (LOADDEDX /
  restricted dE/dx) so the local energy deposit contributes to the DLET/TLET
  numerator/denominator like a track step.
- Handle the zero-track-length case for track-averaged (TLET) vs dose-averaged (DLET).

## Acceptance
- nucre1/nucre2 heavy-recoil **DLET** becomes nonzero and physical
  (`tools/plot_nucre.py` LET-by-species panel).
- The #212 recoil-LET contribution of Stages 1/3 becomes visible.

## Relates
#212 (validation), #179 (score_point).

PR #246, opened 2026-07-05
feat: implement thermal neutron transport model with energy freeze an…

### LINKED ISSUE #178: PHYSICS: Thermal neutron transport: implicit capture + sub-keV cutoff
## Background

The neutron cross-section tables now cover 1 meV – 20 MeV (31-point irregular grid,
3/decade thermal–resonance, 2/decade fast).  The data for B-10(n,α) and Li-6(n,α)
is correctly present down to 1 meV, enabling future BNCT dose and ⁶LiF/⁷LiF TLD
response modelling.

The transport cutoff is currently fixed at **1 keV** (`OSH_NEUTRON_E_CUTOFF_MEV =
1e-3` in `osh_transport_neutron.c`).  Neutrons below this are killed and their
energy deposited locally (KERMA approximation).

## What this issue tracks

Enable explicit thermal neutron transport below 1 keV.  This requires two changes:

### 1. Implicit capture (variance reduction)

At thermal energies H-1 has σ_el/σ_γ ≈ 80/0.33 ≈ 240, so a thermal neutron in
water would take ~240 explicit steps before capture — too expensive for analogue
transport.  Standard solution: **implicit capture**.

At each collision multiply the neutron weight by σ_scatter/σ_tot and always
scatter; skip explicit capture events.  When weight falls below a threshold
(e.g. 10⁻³ of the starting weight), play Russian roulette.  This reduces the
effective step count by ~σ_el/σ_γ without biasing the flux estimate.

Implementation site: `osh_transport_neutron.c`, `apply_event()` / main loop.

### 2. Lower transport cutoff

Drop `OSH_NEUTRON_E_CUTOFF_MEV` from `1e-3` to something around `1e-9` (1 meV,
the lower edge of the XS table).

### 3. S(α,β) caveat

Below ~1 eV, molecular binding in H₂O significantly changes the scattering
kinematics (the free-gas H-1 cross section gives the right magnitude via 1/v but
wrong energy/angle transfer).  For the first implementation it is acceptable to
use the free-gas (0K JEFF) cross sections and note this approximation in the
code and documentation.

## Applications unlocked

- ⁶LiF vs ⁷LiF paired-TLD response (Li-6(n,α), σ_thermal ≈ 940 b)
- BNCT dose from ¹⁰B(n,α) (σ_thermal ≈ 3840 b)
- Thermal neutron fluence downstream of moderating phantoms

## Related

- Closes when implicit capture is implemented and cutoff lowered.
- XS data is already present (branch `154-neutron-transport-minimal-model`).
- New branch recommended; do not mix with the current minimal-model branch.

nbassler:
…d clamping

### DIFF
--- a/src/beam/osh_beam.c
+++ b/src/beam/osh_beam.c
@@ -511,7 +511,11 @@ void osh_beam_print(struct beam_workspace const *wb, struct osh_diag_sink const 
     OSH_DIAG_INFOF(diag, "%-40s : %i", "Random seed offset", wb->rndoffset);
     OSH_DIAG_INFOF(diag, "%s", "");
     OSH_DIAG_INFOF(diag, "%-18s : %f", "DeltaE/E", wb->deltae);
-    OSH_DIAG_INFOF(diag, "%-18s : %f MeV", "Neutron cut", wb->ncut);
+    if (wb->ncut > 0.0f) {
+        OSH_DIAG_INFOF(diag, "%-18s : %g MeV", "Neutron cut", (double) wb->ncut);
+    } else {
+        OSH_DIAG_INFOF(diag, "%-18s : %s", "Neutron cut", "auto (transport default)");
+    }
     OSH_DIAG_INFOF(diag, "%-18s : %f MeV/nucleon", "DeltaE min", wb->demin);
     OSH_DIAG_INFOF(diag, "%s", "");
     OSH_DIAG_INFOF(diag, "%-18s : %s", "Scatter mode", osh_beam_mscat_names[(int) wb->scatter]);
--- a/src/physics/README.md
+++ b/src/physics/README.md
@@ -210,7 +210,10 @@ The current reaction model is intentionally minimal:
 
 - elastic scattering uses an isotropic CM approximation; H-1 recoils are
   returned as proton secondaries, while heavier recoils deposit their recoil
-  energy locally.
+  energy locally.  At or below the thermal energy (`OSH_NEUTRON_THERMAL_E_MEV`,
+  0.0253 eV) the collision is *energy-frozen*: the neutron changes direction only
+  (isotropic in the lab) and keeps its energy, and a down-scatter crossing that
+  threshold is clamped up to it (see the thermal model below).
 - `(n,gamma)` capture kills the neutron and deposits the neutron kinetic energy
   locally.
 - `(n,p)` and `(n,alpha)` use two-body relativistic kinematics when the mass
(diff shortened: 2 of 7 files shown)


nbassler:
Closing: **thermal neutron transport is delivered** — via the *analogue freeze*
model, not implicit capture. (PR #___)

## What landed

- **Sub-keV cutoff lowered.** Default neutron transport cutoff dropped from 1 keV
  to the cross-section table floor, 1 meV (`OSH_NEUTRON_CUTOFF_DEFAULT_MEV` in the
  new `osh_neutron_const.h`), so neutrons slow to thermal by default. `NEUTRLCUT`
  still overrides (e.g. `1e-3` restores the historical 1 keV kill).
- **Thermal transport (one-group freeze).** At/below the thermal energy
  (`OSH_NEUTRON_THERMAL_E_MEV` = 0.0253 eV) elastic collisions only redirect the
  neutron — energy frozen; a down-scatter crossing the threshold is clamped up to
  it. Thermal neutrons then transport at that fixed energy until they **capture or
  escape** the geometry (no arbitrary low-energy kill).
- **S(α,β) caveat documented.** Free-gas 0 K cross sections are used unchanged; the
  freeze avoids the worst artifact (down-scatter toward zero) but not the kinematics.

Evidence (nucre1, 200 MeV p in water): neutron cutoff-kills **161 → 0**, transport
steps **2635 → 10046** (thermal wandering), runtime change negligible. Regression:
running with `NEUTRLCUT 1e-3` reproduces the pre-change output **byte-for-byte**, so
the fast/epithermal path is provably untouched.

## Why implicit capture was deliberately *not* implemented

This issue proposed implicit capture (survival biasing + Russian roulette) to make
analogue thermal transport affordable. On closer look that motivation doesn't hold:

1. **Runtime is a non-issue.** Secondary-neutron yield in hadron therapy is low, so
   analogue thermal transport is cheap (see the negligible runtime delta above).
2. **Implicit capture wouldn't help runtime anyway** — survival biasing keeps a
   neutron alive *longer* (≈ `ln(1/w_min)×` more collisions before roulette) than
   analogue capture terminates it. Its real benefit is **variance reduction of the
   capture reaction-rate tally** for trace nuclides (¹⁰B / ⁶Li → BNCT / TLD).
3. That benefit requires a **capture reaction-rate collision estimator**, and
   capture/recoil dose scoring is **not yet wired** (reaction products are currently
   computed then discarded). So there is nothing for implicit capture to accelerate
   today.

So implicit capture is deferred until (and if) capture-dose scoring lands and
benchmark data justifies the variance-reduction work — at which point it should be
opened fresh with that motivation, not the (incorrect) step-count rationale.

## Deferred follow-ups

- **(n,γ) capture-γ** — `(n,γ)` still deposits the incident KE (≈0 at thermal); the
  ~2.2 MeV H-1 capture photon is not yet modeled.
- **Capture / recoil dose scoring** — via `score_point` / ion-feedback (pre-existing
  gap; relates to #227).
- **Proper thermal scattering** — S(α,β) / free-gas thermal motion. The freeze pins
  all thermal neutrons to a single energy, so a neutron spectrum would show a delta
  line at 0.0253 eV, not a physical Maxwellian.
- **Thermal-neutron benchmark data** — planned as a separate issue.

Issue #230, opened 2026-07-03
parallel: --score-replicas N — sequential private-accumulator + merge harness (deposit-target routing + reproducibility gate)


grzanka:
**Parent:** #161 · **Delivers the deterministic core of** #168 · **De-risks / unblocks** #195, #171

## TL;DR

Add a hidden diagnostic flag `--score-replicas N` that splits the run's `[0, nstat)`
histories into **N contiguous sub-ranges, runs them one after another (no threads)**,
each depositing into its **own private accumulator set**, then **merges all N into the
master** before the normal postprocess + save. This is the first thing in the tree that
actually exercises the private-accumulator + `osh_scoring_accumulator_merge()` path that
every parallel backend (threads, MPI, WASM Web Workers) will depend on — with **zero
concurrency risk**, because the replicas run sequentially.

The value is disproportionate to the effort: it (1) wires the last mile of deposit-target
routing that `--parallel N` (#195) and the WASM workers (#171) both need, (2) proves
`merge == serial` within FP tolerance and is bit-reproducible run-to-run — the acceptance
guarantee the whole parallel design rests on (#168), and (3) is the measurement vehicle for
the #161 profiling study that chooses `--score-mem` defaults, all **before** any threading
backend exists.

> This is deliberately **not** parallel execution. It is the behaviour-preserving refactor +
> test that makes turning on threads a small, well-guarded step instead of broad surgery.

---

## Why now (what exists, what's the gap)

The parallelism groundwork has landed and is in the tree:

- **Work-range loop:** `osh_transport_ion_run_range(...)` transports exactly `[hist_lo, hist_hi)`
  through an `osh_worker_context` (`src/transport/osh_transport_ion.c:442`,
  `src/transport/osh_worker_context.h:43`). RNG is seeded per **global** history index
  (`rndoffset + history`), so any partition of `[0, nstat)` yields disjoint, non-overlapping
  random streams — the scored result is independent of how the range is split, up to FP
  summation order.
- **Deposit-target seam (live):** `osh_scoring_score_step(rt, acc_set, scratch, part, st)` already
  takes a **caller-owned** deposit target + scratch (`src/scoring/runtime/osh_scoring_step.h:37`).
- **Reduce (live):** `osh_scoring_accumulator_merge(dst, src)` is representation-aware — additive
  for `data`/`data2`/`weight`/`nbatch`, Schubert–Gertz for the Welford M2 arrays, additive when
[...]

PR #247, opened 2026-07-05
scoring: report Monte-Carlo standard error via batch-means (#209)

### LINKED ISSUE #209: SCORING: report Monte-Carlo standard error (error bars) on scorer output
## Motivation

OSH reports **no statistical uncertainty** on scored quantities. During the #190 straggling validation we could not separate a genuine scoring artifact from single-particle Poisson noise in the distal LET tail without manual re-runs at higher NSTAT and ad-hoc stop-histograms. We should report a per-bin **standard error of the mean** on every scored quantity and surface it in plots.

## Builds on

- **#169 (closed)** — variance *representation* already decided: Welford/West `(W, M2)` + Schubert–Gertz parallel merge. Already scaffolded in `osh_scoring_accumulator` (`data_var`/`data2_var`, `weight`, `nbatch`) and implemented in `osh_scoring_accumulator_merge()`. Representation is settled — this issue is about **wiring** it end-to-end.
- **#207 (prerequisite)** — batch-aware family-exact scheduler + `osh_transport_ion_run_range([hist_lo,hist_hi))` + `osh_simulation_set_checkpoint_policy(...)`. Variance-batch folding was explicitly deferred out of #207 to a child issue (this one). The M2 fold hangs off #207's per-batch barrier.
- **#193** — activates the currently-dead `beam->nsave` (`NSTAT n step` second field) as the count/dump cadence via `run_ctl_should_dump` + `osh_scoring_snapshot_save`. This delivers the batch boundaries.

## Estimator

Batch-means. A batch = one save-interval block of primaries. Per bin:

```
SE = sqrt( M2 / ((B - 1) · W) )
```

with `B` = number of batches, `W` = Σ history weight, `M2` = Welford sum-of-squared-deviations combined across batches with the Schubert–Gertz cross-term (already implemented). **B = 1 (no save interval) ⇒ 0 d.o.f. ⇒ no error reported** (documented behaviour).

## Decisions already taken

- **Batch source = NSTAT save interval (`nsave`)**, i.e. the #193 count cadence. `B = ceil(nstat / nsave)`.
- **Output = inline paired columns** in the TEXT writer: `... DOSE DOSE_ERR FLUENCE FLU_ERR DLET DLET_ERR TLET TLET_ERR`.
- **DLET/TLET get error bars too** (the LET tail is the whole motivation), not just additive Dose/Fluence — propagate the ratio SE from the two M2 arrays (`data_var` for Σ(L·w), `data2_var` for Σw).

## Scope / task list

- [ ] Enabling card (see open question below).
- [ ] Extend `osh_scoring_accumulator_alloc()` to allocate `data_var`/`data2_var`; set the `page->variance` flag in compile.
- [ ] Fold each batch into master at the checkpoint via `osh_scoring_accumulator_merge()` (needs transport to score into a supplied accumulator set instead of hardcoding the master in `ion_step_commit()`; non-variance path stays byte-identical, zero overhead).
- [ ] Finalize per bin to SE in `osh_scoring_postprocess.c` (ratio propagation for AVER pages).
- [ ] Emit inline `_ERR` columns in the ASCII writer (drop the current variance-page rejection at `osh_scoring_save_ascii.c:420`); BDO error field as a follow-up (`osh_scoring_save_bdo2019.c:343`).
- [ ] Draw error bars in `tools/plot_straggling.py`.
- [ ] Add a 2-batch M2 unit test — the merge cross-term is currently **untested** (`test_osh_scoring_accumulator_merge.c:293`: alloc can't produce var arrays today).

## Stub audit (dormant scaffolding to wire, not delete)

1. `osh_scoring_accumulator_alloc(acc, len, want_data2)` never allocates the M2 arrays — the central gap. Call sites: `osh_scoring_compile.c:944` + test helpers.
2. `page->variance` is a dead flag, read only as a rejection in the two writers.
3. `weight`/`nbatch` are always 0 today — populated once batch folding runs.
4. `merge_m2()`/`variance_inconsistent()` are implemented but the M2 cross-term is untested.
5. SE can be finalized in place into `data_var` (spent after finalize) — no new struct field needed.

Net: one alloc signature extension, two writer-guard flips, one new test; no struct field removals.

## Open question for discussion 👇

Where should the enabling card live, and what is its syntax?

- **(A) Global `detect.dat` card**, e.g. `VARIANCE <on|nbatch>` — matches that batch boundaries are inherently run-global (all pages share the same history-range boundaries).
- **(B) Per-page `Settings` card**, e.g. `Variance <...>` inside each `Settings` block — consistent with the existing per-page overrides (Rescale/Medium/Density), though batch boundaries are still global so it acts as an on/off per output.

Preference / naming / operand semantics? cc @nbassler

grzanka:
## Why

OSH reported **no statistical uncertainty** on scored quantities. During the #190 straggling validation this made it impossible to separate single-particle Poisson noise in the distal LET tail from a genuine scoring artifact. This wires the batch-means variance path end-to-end so every scored quantity carries a per-bin **standard error**.

Closes #209.

## What already existed (this just connects it)

- **#169** — variance representation decided and implemented: Welford/West `(W, M2)` + Schubert–Gertz parallel merge in `osh_scoring_accumulator_merge()`. It was scaffolded but **unwired** (the M2 arrays were never allocated, `weight`/`nbatch` never set).
- **#230** — deposit-target routing (`osh_score_target`), so transport can score into a private accumulator set and merge into the master.

## Changes

- **Enable card**: a global `VARIANCE [N]` card in `detect.dat` turns on tracking for every page (`N` = number of batch-means observations, default 10). `osh_scoring_accumulator_alloc_variance()` allocates the companion M2 arrays. Off ⇒ accumulators and the hot path are **byte-for-byte unchanged**.
- **Per-batch fold** (transport): each checkpoint batch is scored into a private set and folded into the master as one weighted observation (`weight` = primaries in the batch, `nbatch = 1`), so the master's M2 captures the between-batch spread.
  - `run_master_batched` grows **one** reusable batch set (variance on only, off the hot path per §10).
  - `run_score_replicas` tags each replica, so `--score-replicas N` doubles as an N-batch, dump-free variance source.
  - When no cadence is set, the run derives a count cadence that yields `N` batches, so a plain single-threaded run produces error bars out of the box.
- **Finalize** (`osh_scoring_finalize_errors`, before postprocess): converts M2 into a per-bin **relative** standard error — `SE = sqrt(M2 / ((B-1)·W))`; the two-pass AVER quantities (DLET/TLET/Qeff) combine numerator and denominator in quadrature (conservative for the strongly-correlated LET). Stored in `data_var`, so it is invariant under the per-primary / unit scaling the writer applies. `B < 2` ⇒ zero d.o.f. ⇒ no error.
- **Output**: the ASCII writer emits a paired `NAME_ERR` column per quantity (`value × relative-error` = absolute error in the reported units). BDO tolerates variance pages (values only; its error field is a follow-up). Dump shadows clear the variance aliases so a mid-run preview never emits or corrupts the still-live M2.
- Memory estimate accounts for the extra M2 arrays.

## Reproducibility contract

| Property | Holds |
|---|---|
| `VARIANCE N` run twice → byte-identical | ✅ (fixed count-cadence partition is deterministic) |
| Value columns vs a plain run | ✅ within FP tolerance (batching only reorders summation) |
| `B < 2` → error column | all zeros (documented) |

## Tests

- Unit: real-`alloc` 2-batch M2 merge (closes the previously hand-built gap); the finalize estimator (additive, ratio, `B<2`, empty bin, variance-off).
- Integration: a `cases::variance` driver (EM-only + nuclear) asserting value agreement with a plain run, sane non-negative not-all-zero error columns, and byte-identical reproducibility across two runs.
- `plot_straggling.py` shades the error band when present.

Full `ctest` green (debug); scoring/transport/variance suites clean under the ASan+UBSan preset as CI runs it (`detect_leaks=0`); `clang-format` + `clang-tidy` clean on the diff.

## Out of scope (follow-ups)

- BDO standard-error field (values are written; error field deferred, as noted in #209).
- Dumps carry values only (previews); final error bars come from the end-of-run save.

---
_Generated by [Claude Code](https://claude.ai/code/session_01SRVPov7vWGKsd5zgzGqRgv)_


grzanka:
## Cross-check: batch-means SE vs independent-run SE vs SH12A

**Setup:**
- 30 independent OSH runs, N=200 000 each, seeds 89736502–89736531
- 1 OSH batch-means run: N=200 000, `--score-replicas 30`, `VARIANCE 30`
- 30 independent SH12A v1.1.1 runs, N=50 000 each, same seeds
- Beam: ¹²C, 400 MeV/u, pencil beam; water cylinder R=10 cm, L=40 cm; 200 z-bins

---

### Result 1 — OSH batch-means SE ≈ OSH empirical std (main PR #247 check) ✅

| Region | se_batch / std_osh | σ from 1.0 |
|--------|-------------------|------------|
| Full physical region (139 bins, z < 28 cm) | **0.885 ± 0.075** | 1.5 σ |
| Bragg peak zone 20–30 cm (39 bins) | **0.883 ± 0.079** | 1.5 σ |

Expected sampling noise for 30 realisations is ≈ 13% (`1/√(2×29)`).  
A 1.5 σ deviation is comfortably within that window — **the batch-means estimator correctly reports the SE of a single run's mean.**

---

### Result 2 — OSH vs SH12A dose profiles (physics cross-check)

OSH vs SH12A mean dose comparison

| Region | \|OSH − SH12A\| / SH12A |
|--------|--------------------------|
| Entrance z < 2 cm | 5% |
| Plateau z = 2–22 cm | 37% (grows with depth) |
| Bragg peak z = 22–28 cm | 58% |
| Fragmentation tail z > 28 cm | SH12A: ~100 mMeV/g, OSH: 0 |

**The plateau difference is a pre-existing physics gap, not introduced by this PR.**  
SH12A (NUCRE=1) scores dose from _all_ nuclear-reaction products (secondary protons, alphas, heavy fragments) that compensate for primary-beam attenuation and keep the dose plateau flat. OSH currently lacks full secondary-particle dose accumulation, so the plateau declines as primaries are removed by nuclear reactions. The **Bragg peak position** (≈ 27.4 cm) agrees between both codes, confirming the primary-range model is correct.

---

### Result 3 — SE shape comparison (N-normalised)

SE comparison: OSH batch-means, OSH empirical, SH12A empirical

The relative SE (`std/mean`, N-independent) is larger in OSH than in SH12A throughout the plateau because OSH records lower total dose per bin (fewer contributing particles). This is consistent with the physics difference above and does not indicate a problem with the uncertainty estimator.

---

**Summary:** PR #247's batch-means SE implementation is statistically validated against 30 independent runs. The observed ratio se_batch/std_osh = 0.89 ± 0.08 is consistent with 1.0 within the expected Monte-Carlo sampling noise.

cross_check/plots/04_osh_vs_sh12a_mean.png
nbassler, line 1: not sure where the right figure came from, but I am very much against normalizing DDCs like this, since this is very misleading. It hides absolute drift, and factoring that out makes you try to match a curve shou probably shoudnt match the first place. Lets always compare normalized per primary. (NUCRE = 0 should match btw)

docs/user/detect.dat.md
```diff
@@ -208,3 +208,48 @@ Output
 
 `Dose inWater` and `DoseGy inWater` score using the stopping power of water
 regardless of the actual traversed material, equivalent to SH12A dose-to-water.
+
+## Statistical uncertainty (error bars)
+
+A global `VARIANCE` card turns on a per-bin Monte-Carlo **standard error** for
```
nbassler: Should it be global? How about making it a setting cards, which then can be attached to estimators.

src/scoring/runtime/osh_scoring_compile.c
```diff
@@ -216,7 +217,9 @@ enum osh_status osh_scoring_estimate_memory(struct osh_scoring_workspace const *
         for (j = 0u; j < output->npages; ++j) {
             struct osh_scoring_page_def const *page = &output->pages[j];
             enum osh_scoring_score_kind const kind = quantity_to_score_kind(page->quantity);
-            unsigned const arrays = score_kind_uses_data2(kind) ? 2u : 1u;
+            unsigned const sum_arrays = score_kind_uses_data2(kind) ? 2u : 1u;
```
nbassler: Rule §7.2 violation.

src/scoring/runtime/osh_scoring_compile.c
```diff
@@ -926,6 +931,11 @@ enum osh_status osh_scoring_compile(struct osh_scoring_workspace const *ws,
         dst_page->has_data2 = score_kind_uses_data2(dst_page->score_kind);
         dst_page->divide = dst_page->has_data2;
         dst_page->postproc = score_kind_postproc(dst_page->score_kind, dst_page->divide);
+        /* Monte-Carlo standard-error tracking (issue #209): a global detect.dat
+         * VARIANCE card turns it on for every page.  The batch-means M2 arrays are
+         * allocated below; the per-batch fold and finalize happen in transport and
+         * postprocess.  0 keeps the accumulator exactly as before (no var arrays). */
+        dst_page->variance = ws->variance ? 1 : 0;
```
nbassler: §7.2

src/scoring/runtime/osh_scoring_runtime.h
```diff
@@ -168,6 +168,23 @@ static inline struct osh_scoring_scratch *osh_score_target_scratch(struct osh_sc
...
+ * allocated the M2 arrays on every page (issue #209).  Variance is all-or-nothing
+ * (a run-global card), so the first page is representative.  The transport driver
+ * uses this once at setup to decide whether to score each checkpoint batch into a
+ * private set and fold it (with the Schubert–Gertz cross-term) instead of
+ * depositing cumulatively into the master; the finalize/save layers use it to emit
+ * the standard-error columns.  Off ⇒ the accumulators and the hot path are exactly
+ * as before.
+ */
+static inline int osh_scoring_runtime_tracks_variance(struct osh_scoring_runtime const *rt) {
+    return (rt && rt->npages > 0u && rt->pages && rt->pages[0].acc.data_var != NULL) ? 1 : 0;
```
nbassler: §7.2

src/scoring/save/osh_scoring_save_ascii.c
```diff
@@ -354,20 +362,50 @@ enum osh_status osh_scoring_save_ascii_output(struct osh_scoring_workspace const
...
-static void
-fprint_quantity_names(FILE *fp, struct osh_scoring_runtime const *rt, struct osh_scoring_output_runtime const *out) {
+static void fprint_quantity_names(FILE *fp,
+                                  struct osh_scoring_runtime const *rt,
+                                  struct osh_scoring_output_runtime const *out,
+                                  int with_err) {
     size_t ip;
     size_t page_idx;
     char const *qty;
     char const *c;
```
nbassler: a few more inline comments what the varilable hold is always helpful, it makes code scanning quicker for humans

nbassler (commented):
OK i think, we are still WIP, so no PR has to be perfect, it must just state an improvement.

cross_check/plots/04_osh_vs_sh12a_mean.png
grzanka: normalization hallucinated by AI, but the aim was to compare stddev, not the mean values

docs/user/detect.dat.md
grzanka: Agreed — the global `VARIANCE` card is the odd one out next to the per-scorer `Material` / `Density` / LET-axis overrides, and moving it to the attachable "settings card" model is the right direction. One subtlety shapes the exact design, worth writing down before we implement it.

**`VARIANCE` is really two knobs:**

1. **enable (on/off)** — whether *this estimator* allocates the Welford M2 companion arrays and emits the paired `_ERR` columns. Naturally **per-estimator**.
2. **batch count `N`** — how the *whole run* is sliced into batch-means observations. Every estimator shares one partition of the run's histories (each checkpoint batch scores all pages and folds once), so `N` is naturally **run-global**.

So the clean shape is **global default + per-estimator override on the *enable* knob, with `N` staying run-global**. Your settings-card idea maps straight onto it:

```text
VARIANCE 20                # run-global: 20 batch-means observations, ON by default everywhere

Settings
    Name noerr
    Variance Off

Output
    Filename dose_xy.dat
    Geo Map2D
    Quantity Dose
    Settings noerr         # this estimator opts out
```

(or a lighter per-`Output` key — `Variance Off` — if we'd rather not add an `Output`-level `Settings` hook.)

**Why it matters — the 2D-vs-1D split you hit on day one.**
Enabling variance ~doubles an estimator's accumulator memory (each `data`/`data2` gets an M2 twin, in both the master *and* the per-batch private set) and, in `TEXT` output, doubles its column count.

```text
VARIANCE 20

# 1D depth-dose — 300 bins. The error band is the whole point; cost is nothing.
Geometry Cyl
    Name Depth
    R 0 5 1
    Z 0 30 300
Output
    Filename bragg.dat
    FileFormat TEXT
    Geo Depth
    Quantity Dose          # -> DOSE  DOSE_ERR
    Quantity DLET          # -> DLET  DLET_ERR

# 2D dose map — 500x500 = 250k bins. Don't want M2 twins or _ERR on every row.
Geometry Mesh
    Name Map2D
    X -10 10 500
    Y -10 10 500
    Z   0 30   1
Output
    Filename dose_xy.dat
    FileFormat TEXT
    Geo Map2D
    Quantity Dose
    Variance Off           # skip ~4 MB of M2 and the column-doubling over 250k rows
```

| Estimator | bins | extra M2 memory when ON | `TEXT` columns |
|---|---|---|---|
| `bragg.dat` (1D) | 300 | a few KB — negligible | 2 → 4, fine |
| `dose_xy.dat` (2D) | 250 000 | ≈ 4 MB (master + batch set) | doubles, × 250k rows |

The 1D curve is exactly where you want the uncertainty; the 2D map is where doubling memory and file size for noise nobody reads is pure overhead. Per-estimator enable lets you pay only where it helps.

`N` stays on the global card because there is a single run partition — a per-estimator `N` would need several simultaneous partitions (or "finest-`N` + sub-folding" bookkeeping) for no real physical gain.

Good news for scoping: the runtime is already almost there — `page->variance` is per-page, allocation keys off it per page, and `finalize_errors` / the ASCII writer / the accumulator merge all already skip pages without M2. The only all-or-nothing spot is the `tracks_variance()` helper (it samples page 0), which would become "any page tracks variance". So it is a contained change: per-`Output` (or per-`Settings`) enable parse + propagation, that helper, docs, and a mixed-mode test.

Keeping the global card as-is in this PR and tracking the attachable per-estimator design in a separate issue.

---

nbassler:
Some misunderstanding how Settings is used:

from SH12A manual:

image


image

An example using rescale (even if uneccessary since DoseGy does the same), multiplying with
total number of primaries to get absolute dose, and score dose-to-water.
```
Settings
    Name AbsoluteDoseToWater
    Rescale 1.602e-10  # MeV/g -> J/kg
    Primaries 1.34e9  # total number of primaries
    Medium 3  # Assuming that material 3 in mat.dat is water

Output
    Filename NB_dose_1Gy.dat
    Fileformat Text
    Geo MyMesh
    Quantity Dose  # in [MeV/g/primary]
    Quantity DoseGy  # in [Gy/primary]
    Quantity Dose AbsoluteDoseToWater  # [Gy]
```

nbassler:
regarding the global key: instead of introducing global keywords in detect.dat could thay maybe be a commandline option instead?

grzanka:
Replying to the two follow-ups:

**"instead of introducing global keywords in detect.dat could they maybe be a commandline option instead?"** — agreed, and it fits an existing precedent cleanly: `--score-replicas N` is already a CLI-only run parameter (no `detect.dat` card at all), and `--dump-every` / `--dump-every-primaries` already **override** a `beam.dat` card when both are present (`docs/user/command-line.md`: *"A command-line flag always overrides the corresponding beam.dat card"*). A `--variance [N]` flag would follow the same shape: parsed in `src/cli/osh_cli.c`, threaded through `main.c` → `osh_run.c` like `--score-replicas` is today. The one wrinkle vs. `--score-replicas`: `VARIANCE` affects *compile-time* accumulator layout (whether M2 arrays get allocated at all), not just a transport-time parameter, so the CLI value needs to reach `ws->variance` before `osh_scoring_compile()` runs rather than being consumed purely at simulation time — small plumbing, not a redesign.

**Settings usage** — thanks for the correction, confirmed in the parser: a `Quantity` line already accepts multiple trailing names (`output_quantity()` in `osh_scoring_parse_output.c`), each resolved against `Filter`s then `Settings` at compile time (`find_filter_index` / `find_settings_index` in `osh_scoring_compile.c`), and merged per-page via `apply_settings_to_override()`. So the per-estimator VARIANCE toggle should attach the same way your `AbsoluteDoseToWater` example does — `Quantity Dose noerr`, not a separate line — reusing 100% existing grammar/resolution instead of adding a new attachment point.

That also answers where "per-estimator" should live precisely: it's per-`Quantity`-line (= per accumulator page), not per-`Output` and not per-`Geometry`. I checked whether `Geometry` (the closest thing to "detector" as a standalone concept) could carry the toggle instead — it can't cleanly, since one `Geometry` can back multiple independent `Output`s (`compile.c` resolves `output->geometry_name` per output with no dedup), each with its own accumulator. Tying variance to `Geometry` would silently couple unrelated accumulators that just happen to share spatial binning; the per-`Quantity`-line hook sits exactly at the accumulator/memory boundary instead.

So: CLI `--variance [N]` for the global batch-count + default, `Settings`-on-`Quantity-line` for the per-estimator opt-out/opt-in — both reusing existing patterns, no new mechanism needed on either side. Tracking this as follow-up work; keeping the global `detect.dat` card as-is in this PR.

---

nbassler:
P.s.: reason for the design with `Settings` was to keep the format flexible for future extensions, without cluttering the UI with hard-to-read options and used up feature slots (c.f. FLUKA inp), and still retain backwards compability.

nbassler:
everything nice, but would really prefer not to indroduce the global card. It breaks the core design for detect.dat and I fear it may stay there forever.

grzanka:
> everything nice, but would really prefer not to indroduce the global card. It breaks the core design for detect.dat and I fear it may stay there forever.

So you opt for removing the VARIANCE flag from detect.dat and keep "score variance for all quantities by default" policy ? Then number of batches to be controlled only from the command line, without an option to specify it in the input files ?

All the other parameters controlling the run (like number of primaries, how often dump partial results) etc could be controlled both from input files and command line, so why doing differently here ?

Maybe the number of batches should go to `beam.dat` ?

nbassler:
just the global stuff I try to avoid, i really tried hard to keep detect.dat clean, so it doesnt evolve into the messy style the other input files have. (I also dont like mat.dat)

Alternatively if we feel this is necessary, then we should agree there is also a global block, and we should have then a the parser prepared like geoemtry / filters / settings / estimator which can be extended for future global keys. I protest a bit against adding half-baked features into an otherwise rather clean design. They tend to stay forever....

nbassler:
i dont think all CLI optiions exist as keys either, but ok, things go really fast. 
I think it is also worthwhile to think the situation when you want to tune it. Very often for debugging? Then maybe rather CLI than in a file you must edit?

grzanka:
## Decision on `VARIANCE` enablement (after discussion with Niels)

Following up on the design thread above, here's where we've landed. This updates my earlier comments — the global card is *not* being kept as-is; the plan below is what should ship in **this PR**:

1. **Disabled by default.** Variance/standard-error tracking stays off unless explicitly requested — no implicit "error bars everywhere."
2. **No global `detect.dat` card, for now.** The current global `VARIANCE [N]` card is being dropped from this design. Whether (and how) a run-wide default/batch-count should be exposed — CLI flag, some other mechanism, or not at all — needs more thought, so we're deliberately leaving it out rather than shipping something we're not confident in.
3. **Enable per estimator via `Settings`.** Instead, variance tracking turns on per scoring page by attaching a `Settings` block on the `Quantity` line — the same mechanism `Quantity Dose inWater` already uses:

   ```text
   Settings
       Name withErr
       Variance On

   Output
       Filename bragg.dat
       Geo Depth
       Quantity Dose withErr
   ```

I'll implement this myself later — recording the direction here first so it's tracked on the PR before the code changes land.

---

grzanka:
## Implemented the per-estimator `Variance` design decision (comment above)

Pushed two commits to this branch implementing the landing decision from the discussion with Niels:

**`13526d9` — scoring: enable variance per-estimator via Settings, drop global VARIANCE card (#209)**

- **Off by default**, no global `detect.dat` card. Variance tracking is now enabled **per estimator** by attaching a `Settings` block carrying `Variance On` to a `Quantity` line — the exact same mechanism `Quantity Dose inWater` already uses:
  ```text
  Settings
      Name withErr
      Variance On

  Output
      Filename bragg.dat
      Geo Depth
      Quantity Dose withErr
  ```
- The per-page plumbing (accumulator M2 arrays, finalize, ASCII `_ERR` columns) already existed and was already per-page, so this is mostly a rewire:
  - New `Variance On|Off` Settings key (parse + runtime structs).
  - `osh_scoring_compile()` now resolves `page->variance` from the page's own Settings references, before the accumulator is allocated (ordering matters — the M2 arrays are sized at allocation time).
  - `osh_scoring_runtime_tracks_variance()` now ORs over **all** pages instead of assuming page 0 is representative, so mixed on/off pages in the same run work correctly.
  - Batch-count exposure (CLI flag or otherwise) is deliberately **not** added — kept as an internal default (10) that drives the auto-cadence for single-threaded runs, per the "leave it out for now" decision. `--score-replicas N` still works as an explicit batch source.
- Rewrote the unit test and the `run_variance.cmake` / `cases::variance::*` integration tests to use the Settings-based enablement instead of appending a card.
- Updated `detect.dat.md`, `command-line.md`, dev `scoring.md`, and stale code comments to describe per-estimator enablement.

**`4b26a4e` — merge `main`**

Rebased on top of the 5 commits `main` had gained meanwhile (including the `osh_transport_run_minimal` → `osh_transport_run` rename). One real conflict in `osh_transport.c` — main only added Doxygen comments to lines this PR's variance work also touches — resolved by keeping the variance logic and adopting main's comment style.

**Verification**
- `ctest --test-dir build_debug`: 131/131 passed.
- ASan+UBSan preset: scoring/variance/transport suites 22/22 passed.
- `clang-format` clean on the diff.

---

grzanka:
@copilot resolve the merge conflicts in this pull request

PR #249, opened 2026-07-06
245 scoring add usable zone scoring geometry

### LINKED ISSUE #245: SCORING: add usable Zone scoring geometry
## Summary

Add usable `Geometry Zone` scoring to OpenShieldHIT.

The scoring model already contains partial `Zone` support, but it is not enabled end-to-end:

- `include/openshieldhit/scoring.h` documents `Geometry Zone` fields (`zone_start`, `zone_stop`).
- `src/scoring/runtime/osh_scoring_compile.c` maps `kind == "zone"` to `OSH_SCORING_GEO_ZONE` and computes the number of bins from the zone range.
- `tests/unit/test_osh_scoring_parse_geometry.c` covers parsing the current `zones <start> <stop>` geometry line.

However, the runtime scoring and output paths still only support `Mesh` and `Cyl`, so `Zone` cannot currently be used as a real detector geometry.

## Current gaps

- `runtime_supports_geometry()` accepts only `OSH_SCORING_GEO_MESH` and `OSH_SCORING_GEO_CYL`.
- `osh_scoring_score_step()` treats every non-CYL geometry as Mesh and runs spatial raytracing, which is wrong for Zone scoring.
- `osh_scoring_score_point()` should be checked and wired consistently for point deposits.
- ASCII output rejects non-Mesh/non-Cyl geometries.
- BDO2019 output rejects non-Mesh/non-Cyl geometries, even though the legacy geometry-kind mapping already contains `ZONE`.
- User documentation only describes Mesh and Cyl, which matches the currently usable feature set but not the partial data model.

## Proposed semantics

Implement `Geometry Zone` as scoring by transport/GEMCA zone identity, but expose zones to users by **zone name**, not by internal numeric ID. Numeric zone indices are an implementation detail and are not generally visible or stable enough for input files.

Candidate syntax:

```text
Geometry Zone
    Name MyZones
    Zone WaterBox
    Zone Target

Output
    Filename zone_dose.dat
    FileFormat TEXT
    Geo MyZones
    Quantity Energy
```

This detector has one spatial bin per explicitly listed zone.

There is also a SHIELD-HIT/FLUKA-like range question to settle:

```text
Geometry Zone
    Name MyZoneRange
    Zones Box Boxbox
```

This could mean either:

- only the two named zones `Box` and `Boxbox`; or
- the inclusive range of zones defined between `Box` and `Boxbox` in `geo.dat`.

The second interpretation may match legacy/FLUKA-style range semantics, but it is surprising enough that it must be explicitly documented if supported. Prefer a spelling that makes the distinction unambiguous, for example:

```text
    Zone Box
    Zone Boxbox
```

for an explicit list, and:

```text
    ZoneRange Box Boxbox
```

for the inclusive definition-order range.

Important design decisions:

- Resolve zone names against the parsed `geo.dat` zone table during simulation/scoring setup, then store dense internal indices for fast scoring.
- Decide whether to support explicit lists only in the first milestone, or both explicit lists and definition-order ranges.
- Decide and document how out-of-range zones and `OSH_ZONE_INDEX_INVALID` are handled. Likely: score nothing.
- Decide volume/normalisation semantics for zone quantities that need volume or density-like normalisation. `Energy` is straightforward; `Dose`/`DoseGy` need a clear mass/volume model or may need to be limited initially.
- Clarify behavior for voxel/CT zones, where the transport zone can refer to a CT body while material comes from HU lookup.

## Suggested implementation plan

1. Parser/docs
   - Document `Geometry Zone` in `docs/user/detect.dat.md`.
   - Add parser support for name-based zone selection.
   - Prefer explicit `Zone <name>` entries first; add `ZoneRange <first> <last>` only if we want legacy range behavior.
   - Add parser tests for explicit zone lists and, if supported, definition-order ranges.

2. Compile/runtime enablement
   - Resolve zone names to internal GEMCA/transport zone indices after `geo.dat` has been parsed.
   - Allow `OSH_SCORING_GEO_ZONE` in `runtime_supports_geometry()`.
   - Size accumulators from the resolved zone list length, not from user-visible numeric IDs.
   - Ensure Zone geometries do not require raytrace crossing scratch.

3. Step scoring
   - Add a Zone-specific branch in `osh_scoring_score_step()`.
   - Use the current transport step zone (`st->zone`) to select the detector bin through the resolved zone-index mapping.
   - For track-length quantities, deposit the whole step contribution into that zone bin. If a step can cross a zone boundary, confirm transport already splits at geometry boundaries before scoring; otherwise document/handle the approximation.

4. Point scoring
   - Wire `osh_scoring_score_point()` so point-deposited energy lands in the current zone bin when the scorer geometry is Zone.

5. Output
   - Add ASCII output for Zone, with a clear zone-name column.
   - Add BDO2019 Zone output if the existing legacy `ZONE` metadata can represent the selected zones; otherwise document TEXT-only as the first milestone.

6. Tests
   - Unit test zone-name resolution: first zone, last zone, duplicate/unknown names, invalid zone.
   - If `ZoneRange` is supported, test inclusive definition-order expansion.
   - End-to-end case with a simple multi-zone geometry and Zone TEXT output.
   - Include at least `Energy` initially; add `Dose`/`DoseGy` once mass/normalisation semantics are settled.

## Acceptance criteria

- A name-based `Geometry Zone` detector can be parsed, compiled, scored, and saved.
- Zone scoring does not expose internal numeric IDs in user input.
- Zone scoring does not go through Mesh/Cyl raytracing.
- Out-of-range and invalid zones are handled deterministically.
- At least one end-to-end CTest case validates Zone output.
- `docs/user/detect.dat.md` documents the supported syntax, quantities, and limitations, including the difference between explicit zone lists and any supported range syntax.

nbassler:
This pull request introduces a major redesign of how Zone-based scoring geometries are handled, moving from index-based zone selection to explicit, user-named zone selection with per-zone volume support. It also adds robust parsing, validation, and error handling for these new Zone geometry features, and updates both the documentation and code to reflect the new workflow. The most important changes are grouped below.

**Zone Geometry Redesign and Parsing:**

* Zone scoring geometries are now defined by explicit `Zone <name>` selectors (matching names in `geo.dat`), with an optional `Volume <cm3>` for each zone. The parser stores zone names and volumes, and the mapping to transport zone indices is resolved at the application level after both geometry and scoring files are loaded. (`src/apps/osh/osh_scoring_parse_geometry.c`, `src/apps/osh/osh_scoring_parse_keys.h`, `include/openshieldhit/scoring.h`, `docs/user/detect.dat.md`, [[1]](diffhunk://#diff-e791f67a8575f50d3329c926d966846541332c2e52b115ef2c4c6a68da032833R269-R332) [[2]](diffhunk://#diff-1a70193b262d2086fc586685cb2a1672fb0bf325dedca0ed09ae79098d885875L11-R12) [[3]](diffhunk://#diff-1a70193b262d2086fc586685cb2a1672fb0bf325dedca0ed09ae79098d885875R39) [[4]](diffhunk://#diff-1a70193b262d2086fc586685cb2a1672fb0bf325dedca0ed09ae79098d885875L56-R64) [[5]](diffhunk://#diff-1a70193b262d2086fc586685cb2a1672fb0bf325dedca0ed09ae79098d885875L82-R91) [[6]](diffhunk://#diff-1a70193b262d2086fc586685cb2a1672fb0bf325dedca0ed09ae79098d885875R145-R177) [[7]](diffhunk://#diff-1a70193b262d2086fc586685cb2a1672fb0bf325dedca0ed09ae79098d885875L216-R301) [[8]](diffhunk://#diff-d92303969ad6f13471429702279f3a9474675b84cee7739e4dcc0bf18dd8ad0fL33-R34) [[9]](diffhunk://#diff-e8aa84da844a4e5d7724b3473e988dfc03360b64eab42068f7ac1d82636b0fd6L203-R203) [[10]](diffhunk://#diff-e8aa84da844a4e5d7724b3473e988dfc03360b64eab42068f7ac1d82636b0fd6L215-R219) [[11]](diffhunk://#diff-e8aa84da844a4e5d7724b3473e988dfc03360b64eab42068f7ac1d82636b0fd6L234-R237) [[12]](diffhunk://#diff-8b72bf318e2cece1020c01d009f9661399bf586defa2120d6156f06d1da8eccdR78-R106)

* Added a new function, `osh_scoring_resolve_zone_names`, which resolves user-specified zone names to transport indices using the loaded geometry, and emits errors for unknown zones and warnings for missing volumes. This function must be called before compiling scoring workspaces. (`src/apps/osh/osh_app_osh.c`, `src/apps/osh/osh_app_osh.h`, `src/apps/osh/osh_run.c`, [[1]](diffhunk://#diff-e791f67a8575f50d3329c926d966846541332c2e52b115ef2c4c6a68da032833R269-R332) [[2]](diffhunk://#diff-2fda8f148d32350fda12efba34da07fc41761557ca9a9a6a6c9fb1ecfeb4cb66R84-R108) [[3]](diffhunk://#diff-559d5132c2ba04439af7100acff3d22eca02e07f1c92ccf0b3557bdf878bdd91R264-R273)

**Documentation Updates:**

* Updated developer and user documentation to describe the new Zone geometry workflow, including how estimators are handled, the role of per-zone volumes, and the expected output formats. (`docs/dev/scoring.md`, `docs/user/detect.dat.md`, [[1]](diffhunk://#diff-7753603ef38a94eed5eb6193f2e4b11944d3115d6bc66e9ba77822da5f0c7711L257-R302) [[2]](diffhunk://#diff-8b72bf318e2cece1020c01d009f9661399bf586defa2120d6156f06d1da8eccdR78-R106)

**Internal API and Data Structure Changes:**

* The `osh_scoring_geometry_def` struct now holds arrays for zone names, resolved indices, and per-zone volumes, replacing the previous start/stop index range approach. (`include/openshieldhit/scoring.h`, [[1]](diffhunk://#diff-e8aa84da844a4e5d7724b3473e988dfc03360b64eab42068f7ac1d82636b0fd6L203-R203) [[2]](diffhunk://#diff-e8aa84da844a4e5d7724b3473e988dfc03360b64eab42068f7ac1d82636b0fd6L215-R219) [[3]](diffhunk://#diff-e8aa84da844a4e5d7724b3473e988dfc03360b64eab42068f7ac1d82636b0fd6L234-R237)
* Updated memory management to correctly free per-zone name strings. (`src/scoring/osh_scoring.c`, [src/scoring/osh_scoring.cR23](diffhunk://#diff-2ceaf222c9d98bcaf54c9c3208641e90ea6e88e746f5102d44b133ef95e1be45R23))

**Build and Runtime Integration:**

* Added new estimator implementation files to the scoring library build. (`src/scoring/CMakeLists.txt`, [src/scoring/CMakeLists.txtR5-R8](diffhunk://#diff-9fd8d8ed1e23a3d8d26162dff96d0e5c866ac62bda558fe0f7b5f742bbb3a3e8R5-R8))

These changes make Zone-based scoring more robust, user-friendly, and extensible, while improving error handling and documentation.


grzanka (commented):
## Architecture review — #249 (usable Zone scoring + estimator/postprocess refactor)

*Reviewer: **Claude Opus 4.8** (Claude Code), reasoning effort **high**. Focus per request: **architecture**, not a bug hunt — the physics/numerics findings in #248 are deliberately out of scope here except where they touch this PR's new code. Findings are model-generated from reading the branch at `89f766a`; please spot-check before acting.*

This PR does **two** things at once: it delivers the Zone-scoring feature from #245, and it lands a sizeable refactor — a per-`Quantity` **estimator registry** plus a move of **÷volume normalisation from score-time into postprocess**. The refactor is the wider-blast-radius change, so most of this review is about it.

### What's good (and should be kept as-is)

- **The geometry ⟂ estimator split is real, not just narrated.** `osh_scoring_score_step()` does one geometry traversal (Mesh/Cyl raytrace, Zone = one bin) and then dispatches each score group through `osh_scoring_estimator_for()` → `score_step`/`score_point`. Geometry decides *which* bins; the estimator decides *what/how*. That is the right axis to cut this module along, and the registry (`osh_scoring_estimator.c`) keeps "add a Quantity = one handler trio + one row" honest.
- **Moving ÷volume into a single `postprocess` seam (`geo->bin_vol_inv`) is the correct call**, and it is *architecturally aligned* with parallelism rather than merely compatible: raw extensive sums (and the two-pass `data`/`data2`) merge **additively**, and the single volume division happens once *after* merge. This is exactly the ordering `run_score_replicas()` already assumes on `main` (`src/transport/osh_transport.c` → merge private accumulators, then `osh_simulation.c:699` postprocess, then save). The `osh_score_deposit()` seam + descriptor/accumulator split remain the right substrate for #161/#169.
- **`docs/dev/scoring.md` is excellent** — the contract table (score_step / score_point / postprocess per Quantity) is precisely the artifact that keeps a registry from rotting. Please keep this rigor.
- **Layering is respected (§9):** zone *names* resolve to dense transport indices at the app layer (`osh_scoring_resolve_zone_names`, cold), the runtime stores dense indices only, and Zone scoring bypasses raytrace — matching #245's acceptance intent.

### Architecture concerns / cleanups

**1. Dead work left on the hot path (should be removed).** Volume normalisation moved to `bin_vol_inv`, but the step driver still *fills* `crossings[j].vol_inv` for every crossing — the Mesh loop (`osh_scoring_step.c:149-151`) and the Cyl LUT loop (`:175-177`) — and **no estimator reads it anymore** (`grep` finds only writes + one stale doc line). `mesh_geometry_to_grid()`'s `voxel_volume_inv` output and its `1.0/(dx*dy*dz)` are likewise dead. This is per-step, per-crossing waste and cuts against the project's own hot-path discipline (§10.3). Also, the comment at `osh_scoring_estimator_step.c:108` ("`vol_inv` is read per-crossing from `crossings[j].vol_inv`") now **documents behaviour that was removed** — it should be corrected before it misleads the next contributor (§4.4).

**2. Three parallel volume representations.** `cyl_vol_inv[r]`, `zone_vol_inv[z]`, and the new fully-expanded `bin_vol_inv[flat]`. `bin_vol_inv` denormalises the other two (Mesh: one scalar stored `nbins` times; Cyl: `nr` expanded to `nr×nz`). That's a deliberate space-for-uniformity trade so postprocess can index `bin_vol_inv[i % stride]` with no geometry switch — reasonable — but after compile the first two arrays are pure build inputs. Consider dropping them from the runtime struct once `bin_vol_inv` is built, or documenting why both are retained.

**3. The four two-pass handlers duplicate their setup.** `score_step_dlet`/`score_step_tlet` are ~30 identical lines of table lookup + Settings-override handling, differing only in the weight (dose vs track); same for `dqeff`/`tqeff`. The per-bin deposit loop is genuinely hot and fine to keep explicit, but the **setup** could collapse into a shared `(value, weight) → (num, den)` helper without adding per-bin indirection. Four near-clones will drift.

**4. Namespacing of the new cross-TU symbols (§6.1).** `score_step_*`, `score_point_*`, `zone_bin_index`, `mesh_geometry_to_grid`, `cyl_geometry_to_grid` are all *external-linkage* symbols with generic, unprefixed names (declared in `*_internal.h`). §6.1 knows only `osh_`-prefixed public API and `static` file-locals — and the sibling helpers in `osh_scoring_estimator_common.c` *do* use `osh_scoring_estimator_*`. These don't. Generic global names like `score_step_energy`/`axis_index` invite link-time collisions; either prefix them (`osh_scoring_estimator_step_energy`) or make static where a single TU suffices.

**5. `NKERMA` is a half-citizen of the registry.** Its row is `{NULL, NULL, postprocess_volume}` — deposited somewhere outside the estimator path but finalised inside it. That's a real asymmetry in the "estimator owns the trio" model; a one-line note in the contract table (where NKERMA's deposit lives) would stop it reading as an omission.

**6. Forward coupling with the #169 variance contract.** Postprocess now transforms `data` (÷volume, and the `data/data2` ratio) but never touches `data_var`/M2. `scoring.md §4` still claims wiring variance later is "a local change" — but because ÷volume now lives *in* postprocess, once M2 is live `postprocess_volume` must scale M2 by `vinv²` and `postprocess_ratio` must propagate error through the quotient. No bug today (`data_var` is NULL), but this PR quietly put the postprocess handlers on the variance-finalisation critical path; the contract in §4/§6 should say so.

**7. Minor:** the Mesh/Cyl/Zone grid builders live in `osh_scoring_step.c` though declared in `..._geometry_runtime_internal.h` (no `geometry_runtime.c` TU) — a reader hunting geometry helpers won't find them in a geometry file. And `osh_scoring_postprocess()` is **single-shot / not idempotent** (a second in-place call double-divides / re-ratios); fine while `osh_simulation` calls it once, but worth a guard or comment as dump/checkpoint (#170) and merge multiply the call sites.

### What's needed for a green light to merge

1. **Rebase onto `main` and prove the replica path.** This branch forked from `07181a9` — it predates #234 (`--score-replicas` merge harness) *and* #239 (ASan CI). `mergeable_state` is "clean" *textually*, but the entire value of the volume-in-postprocess change is that it must compose with merge-then-postprocess. After rebasing, run a DOSE/FLUENCE case with `--score-replicas N` and confirm it equals the serial result (raw sums merge, ÷volume once). This is the single highest-risk item and it's invisible to CI today because the harness isn't in the base.
2. **Add the end-to-end Zone CTest case.** #245 acceptance criterion #5 ("at least one end-to-end CTest case validates Zone output") is **not met** — there is no `tests/cases/*` Zone detector, only unit tests. Add a multi-zone geo + `Geometry Zone` `detect.dat` + TEXT output checked against a reference.
3. **Cover (or gate) the Zone save paths.** Zone ASCII/BDO save tests are explicitly deferred to "Stage C" (`test_osh_scoring_save.c:283`), yet the PR *ships* both writers. Codecov backs this up: `save_ascii` 23%, `save_bdo2019` 21%, `app_osh` 26% patch coverage. Either add smoke tests, or land Zone as TEXT-only first (as #245 permits) and have BDO Zone reject loudly.
4. **Close the ASCII multi-page diff-axis item.** The Copilot thread on `save_ascii.c:500` (header shape taken from `p0`, but the data loop indexes with each page's own `diff_stride`/`diff_nbins`, `:156`) is marked resolved but *not* outdated — confirm a guard was actually added that rejects pages whose diff config differs from `p0`, otherwise a later page can compute an out-of-range `data_idx`.
5. **Clean the history / consider splitting.** 14 commits including `[skip ci]` WIP and "more comments", and two distinct logical changes bundled. Per CONTRIBUTING (one logical change per commit): at minimum squash into coherent commits. Better — land the **estimator/postprocess refactor as its own PR first** (it stands alone and is the risky part), then Zone on top. It would make the refactor reviewable independently and shrink the rebase blast radius.
6. **Remove finding #1's dead code + fix the stale comment** — small, and keeps the hot path honest.

**Bottom line:** the design direction is right and the refactor is a genuine improvement that *helps* the parallelism roadmap rather than fighting it. It's not mergeable yet mainly for process/verification reasons — rebase-and-verify against the replica merge (1), the missing end-to-end Zone test (2), and Zone-save coverage (3) — not for anything structurally wrong.

---

grzanka (approved):

grzanka:
## Cross-check: volume-in-postprocess composes correctly with the replica-merge path

Follow-up to my [architecture review](https://github.com/openshieldhit/openshieldhit/pull/249#pullrequestreview-4635784579). That review flagged one item as the highest merge risk: this PR moves ÷volume normalisation out of score-time and into `postprocess` (via `geo->bin_vol_inv`), but the branch predates #234's `--score-replicas` merge harness — so CI has **never** tested the two together. The concern: if the merge summed raw accumulators while volume division happened per-replica or before the merge, dose/fluence would be silently wrong under any parallel/merged run.

I verified it empirically. **Result: the composition is correct.**

*Verification by **Claude Opus 4.8** (Claude Code), reasoning effort **high**. Reproducible; numbers below.*

### Method
- Merged this PR's current head (`2b4a4a3`) into `main` (`a5988bd`, which contains the #234 harness) → **clean merge, 0 conflicts**.
- Built the debug preset and ran each case twice: a plain serial run vs `--score-replicas N` (which scores N contiguous sub-ranges into private accumulator sets, merges them, then postprocesses). `-n 2000 -N 0` (fixed seed) both runs.
- Compared every value column of the TEXT output.

### Results — serial vs `--score-replicas`, max **relative** difference per column

| Case | Quantities exercised | N | max \|rel\| |
|---|---|---|---|
| `07_settings_filters` (Mesh) | DOSE, DOSEGY, FLUENCE, DLET, TLET, DQEFF, TQEFF — every postprocess path (÷volume, ×Gy, `data/data2` ratio) | 8 | **9.8e-13** |
| ″ | ″ | 16 | **9.8e-13** |
| ″ | ″ | 3 (uneven 667/667/666) | **9.8e-13** |
| `08_minimal_cyl` (Cyl) | per-R-bin volume path | 8 | **1.1e-13** |

That's floating-point summation-reordering noise. A real composition bug (per-replica volume division, division before merge, or double-counting in the two-pass ratio) would show O(1) or factor-of-N discrepancies — not 1e-13. The uneven 3-way split also confirms unequal batch sizes merge correctly.

- **Merged tree builds clean and passes 100% of tests (125/125)** — including `test_osh_score_replicas` and the scoring suites, i.e. the exact PR-refactor + merge-harness combination CI doesn't currently cover.
- The 13 skipped `reference::idd_*` tests are the pre-existing no-reference-curves tier (unrelated to this PR).

### Takeaway
The scariest architectural risk is cleared: raw extensive quantities merge additively across private accumulators, and the single ÷volume (and the ratio finalisation) is applied once in postprocess after merge — identical to serial within FP tolerance. The rebase onto `main` is therefore mechanical and low-risk, not a correctness gamble. Remaining pre-merge items (end-to-end Zone test, Zone-save coverage) and follow-ups are unchanged from the review; a leftover-tasks issue is being drafted separately.

---

PR #266, opened 2026-07-07
Add pre-equilibrium emission model to nuclear physics module

### LINKED ISSUE #225: PHYSICS: pre-equilibrium emission (exciton model) — fast stage part 3/3
**Part 3/3 of the fast nuclear reaction stage (parent: #221). Follows cascade +
coalescence, precedes equilibrium de-excitation.**

## Motivation
Bridge between the fast intranuclear cascade and equilibrium break-up: the residual
still emits fast nucleons/light clusters **before** it thermalizes. Pre-equilibrium
emission carries that intermediate stage and cools the residual before it is handed
to Fermi break-up (A≤16) or SMM/evaporation (A>16).

## Reference
Gudima, Mashnik & Toneev, *Nucl. Phys.* **A401** (1983) 329 (exciton model).

## Scope
- Exciton-model pre-equilibrium emission of nucleons / light clusters.
- Emit as fast secondaries; hand the cooled residual `(A, Z, E*)` to the
  equilibrium stage: A≤16 → Fermi break-up; A>16 → SMM / evaporation-fission
  (see #175 Weisskopf; SMM still TODO).

## Acceptance
Residual excitation and multiplicity are physical for p + ¹⁶O at 200 MeV; the
combined fast-stage chain (cascade + coalescence + pre-equilibrium) leaves an
excited residual consistent with the A≤16 FBU / A>16 branch decision.

### LINKED ISSUE #265: nuclear: absorbed cascade proton keeps energy+momentum in the prefragment but not A/Z (capture bookkeeping)
**Minor bookkeeping inconsistency in the abrasion prefragment, deferred from
#263 to keep that stage's observable attribution clean.**

## Problem

When the cascade proton is absorbed into the prefragment
(`osh_nuclear_abrasion.c`: it cannot pay the per-hole excitation cost, or ends
below ~1 MeV), its **energy and momentum stay with the fragment** (E\* +=
remnant; momentum never subtracted — by design, see the momentum-balance
comment) and since #263 it is counted as a **particle exciton** — but the
fragment's **A and Z are not incremented**. Physically this is a capture: the
residue should become (A+1, Z+1), exactly as the neutron path already builds
its compound nucleus as (Z, A+1) in `osh_neutron_reaction.c`.

Consequence: absorption events hand Fermi break-up a nuclide that is one
proton lighter than the system that actually holds the booked energy and
momentum. Energy/momentum conservation is unaffected (both are exact); only
the A/Z identity of the de-exciting system is off by one proton.

## Why minor

At therapy energies absorption of the cascade proton is rare (~10% of
inelastic events at 200 MeV, dominated by low-T_current chains); the effect
concentrates at low incident energies (end of range), where the FBU products
of neighbouring nuclides are similar. No observable in the #260/#263
benchmark suite is expected to move visibly.

## Fix sketch

In both absorption paths, increment `fragments[0].a` and `fragments[0].z`
(and skip the corresponding decrement bookkeeping); update
`test_retention_low_energy` (A becomes A+1 in full-absorption events) and the
momentum-balance comment. Bundle with Stage 2 (#225) or as a standalone
one-liner PR — either works; do not fold silently into unrelated physics
changes.

## Related

- #263 (where the particle-exciton count was added and this was flagged)
- #221 (fast-stage epic)

### LINKED ISSUE #267: transport: non-default isotopes use the representative isotope's range table and rest mass (deuteron range ~2x wrong)
**Non-default isotopes are transported with the representative isotope's
range table and rest mass — a deuteron's range is wrong by ≈2×.**

## Mechanism (three coupled pieces)

1. Runtime tables have **one projectile column per Z**, built for the default
   isotope (`osh_material_compile.c:698-716`: Z=1→A=1, Z=2→A=4, …), and the
   CSDA range integral bakes that A in (`integrate_range(..., proj.a, ...)`,
   `osh_material_compile.c:161-186`; R(E/u) = ∫ A·d(E/u)/SP — range is
   proportional to A at fixed E/u).
2. Transport maps **any** isotope of charge Z onto that column:
   `find_projectile_index` (`osh_transport_ion_step.c:1258-1261`) explicitly
   ignores an A mismatch ("differing isotopes share the representative
   projectile for that Z for now") — with **no runtime warning**.
3. The per-step kinematics then mixes the particle's A with the column's
   mass: `e0/a_proj` uses the true A for the E/u lookup (good), but
   `ctx->proj_mass_mev = material_rt->projectile_mass_mev[projectile_idx]`
   (`osh_transport_ion_step.c:586`) is the representative isotope's mass.

## Consequences

Mass stopping power at fixed E/u is isotope-independent; **range is not**
(R ∝ A/z²):

- **Deuterons** (Z=1) use the proton range column: true R_d(E/u) = 2·R_p(E/u),
  so a deuteron's sampled energy loss per cm is 2× too high and its range is
  **half** the physical value. Tritons: 3×. ³He (Z=2 column is ⁴He): range
  overestimated by 4/3. (He-4 itself is the representative — unaffected.)
- β²/γ in the straggling parameters (ξ, E_max,
  `osh_transport_ion_step.c:911-914`) and in Highland pv
  (`osh_physics_scat_highland.c:28-33` via `proj_mass_mev`) use the wrong
  rest mass — a deuteron is treated as a proton with twice the kinetic
  energy (γ−1 doubled).

## Who is affected

- d/t/³He secondaries from Fermi break-up and (since #225) pre-equilibrium
  emission: small dose fraction, but their spatial distribution is
  systematically wrong.
- **The #225 / #212 cluster-fluence validation**: the plateau Φd/Φt/Φ³He
  ratios vs SH12A are distorted (halved deuteron track length suppresses Φd
  by roughly ×2), so the pre-equilibrium γ_d calibration must be re-measured
  after this fix.
- Much worse: any **non-default-isotope primary beam** (deuteron or ³He
  beams are legitimate use cases) — the run proceeds with no warning.

## Fix direction

Within a Z column, rescale range lookups by A_actual/A_rep (exact up to the
tiny Tmax term in the Bethe log) and use `part->mass` for kinematics — both
cheap; or add isotope columns for Z = 1, 2. At minimum, WARN once per (Z, A)
when the mismatch branch at `osh_transport_ion_step.c:1258` is taken.

## Related

- #225 (pre-equilibrium d/t/³He production — calibration blocked on this)
- #212 (secondary dose/LET validation vs SH12A)
- #213 / #214 (ion secondary transport)

nbassler:
- Implemented the pre-equilibrium exciton model to handle fast ejectile emissions between abrasion and equilibrium break-up stages.
- Added new source files: osh_nuclear_preeq.c and osh_nuclear_preeq.h for the model's implementation and interface.
- Updated osh_nuclear_handler to integrate the pre-equilibrium model, ensuring proper compilation and stepping of the model.
- Enhanced nucre_scan example to utilize the new pre-equilibrium model.
- Created unit tests for the pre-equilibrium model, covering various scenarios including thermalized fragments, conservation laws, and emission statistics.

### DIFF
--- a/docs/dev/bug-hunts/2026-07-05-deep-audit.md
+++ b/docs/dev/bug-hunts/2026-07-05-deep-audit.md
@@ -14,7 +14,7 @@
 | **Commit audited** | [`e3c619a`][e3c619a] on `main` (2026-07-05) |
 | **Source** | [PR #248][#248] |
 | **Scope** | Test infra · transport hot path & EM physics · scoring · nuclear/neutron transport · RNG & parallel readiness · GEMCA geometry · parsers/IO · architecture |
-| **Follow-up** | 11 suggested issues (1 filed — [#255][#255], fixed by [#257][#257]) — see [§8](#sec-8) |
+| **Follow-up** | 11 suggested issues (2 filed — [#255][#255], fixed by [#257][#257]; [#267][#267], fixed by [`a0f60a1`][a0f60a1]) — see [§8](#sec-8) |
 
 
 Deep audit of `main` (e3c619a) prompted by PR [#239][#239] (sanitizer wiring) and issues
@@ -160,6 +160,13 @@ cutoff kill is the same physical situation without the deposit.
   with the effective cutoff.
 
 ### P-2 (high, high) Non-default isotopes are transported with the representative isotope's range table and rest mass — deuteron range wrong by ≈2× {: #p-2 }
+
+!!! success "Resolved"
+    Filed as [#267][#267] and fixed by [`a0f60a1`][a0f60a1].  Transport still
+    shares stopping-power columns by `Z`, but CSDA range is now scaled by
+    `A_actual/A_table`, isotope-specific kinematics use `part->mass`, and a
+    regression unit covers non-default H/He isotopes in a non-unit-density
+    material.
 
 Mechanism, three coupled pieces:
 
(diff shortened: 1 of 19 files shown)


nbassler:
[stage2_report.pdf](https://github.com/user-attachments/files/29759718/stage2_report.pdf)

Issue #255, opened 2026-07-06
gemca: missing ×2 in quadric linear coefficient (ELLZ/CONE/ELLIPSOID) + cancellation-prone root formula + per-surface distance unit tests (G-1, G-2)


grzanka:
## Summary

[Bug-hunt report, findings G-1 and G-2][g-1]: `_quadratic_solver()` in `src/gemca/osh_gemca2_dist.c` solves `a·t² + b·t + c = 0` expecting the **full** linear coefficient `b`, but three of the five surface-distance call sites pass the **half**-coefficient instead. Every ray intersection against an ellipsoid, elliptic cylinder, or cone (`ELLIPSOID`, `ELLZ`, `CONE` in `geo.dat`) is computed at the wrong distance whenever `b ≠ 0` — i.e. whenever the ray doesn't pass exactly through the surface's axis/center. This transports particles across the wrong geometric boundary: dose gets systematically misplaced and zones are effectively deformed for any geometry that uses these body types.

Separately (G-2), even the *correct* call sites use the textbook `(−b ± √d)/(2a)` root form, which is cancellation-prone exactly in the state every boundary crossing produces (`c ≈ 0` just after a nudge off a surface).

No existing test or example in the tree exercises `ELL`/`TRC`/`REC` bodies, which is why `ctest` is green despite this.

## G-1 (critical, high confidence) — missing factor of 2

`_quadratic_solver(a, b, c)` (`src/gemca/osh_gemca2_dist.c:482-510`) computes `d = b² − 4ac`, roots `(−b ± √d)/(2a)`. The call sites disagree on what `b` means:

| surface | function | line(s) | `b` passed | correct? |
|---|---|---|---|---|
| sphere | `_dist_sphere` | `:353-363` | `2·(cp·p)` | ✔ |
| cylinder (`CYLZ`) | `_dist_cyl` | `:379-390` | `2·(cpx·px + cpy·py)` | ✔ |
| elliptic cylinder (`ELLZ`) | `_dist_elipcyl` | `:405-417` | `cpx·px/ra² + cpy·py/rb²` | ✗ missing ×2 |
| cone (`CONE`) | `_dist_cone` | `:432-446` | `cpx·px + cpy·py − t·cpz` | ✗ missing ×2 |
| ellipsoid (`ELLIPSOID`) | `_dist_ellipsoid` | `:459-470` | `Σ cpᵢ·pᵢ/rᵢ²` | ✗ missing ×2 |

[...]

PR #285, opened 2026-07-08
feat: Implement residual kinetic energy deposit for ions killed at en…

### LINKED ISSUE #279: transport: residual kinetic energy silently deleted when an ion is killed at the energy cutoff
## Summary

Every ion that ranges out inside the geometry has its remaining kinetic energy deleted instead of deposited. `ion_step_setup()` handles the energy-cutoff kill with no scoring call at all:

```c
// src/transport/osh_transport_ion_step.c:557-562
ctx->cutoff = cutoff_total_energy(params, material_rt, ctx->part);
if (ctx->e0 <= ctx->cutoff) {
    pool->e[slot] = 0.0;
    ctx->done = 1;
    return;
}
```

`ctx->e0` — up to `ctx->cutoff` MeV of real kinetic energy — is overwritten with `0.0` and the particle is dropped. No `osh_scoring_score_point()` call, no point deposit, nothing. The energy simply isn't anywhere in the output.

(Verified against `main` @ `0dbc78b`. The audit cited this as `:539-543` against `e3c619a`; the file has since gained ~19 lines above this point from unrelated commits, so line numbers in this issue are current, not the audit's.)

## This is not a rare corner case

The exit-energy calculation clamps a stopping track's exit energy to **exactly** `cutoff`, never below it — across both step-length regimes and again after straggling re-perturbs it:

- `osh_transport_ion_step.c:811-812` — pre-straggling proposed exit energy (physics-limited step)
- `osh_transport_ion_step.c:920-923` — exit energy before straggling (boundary/CSDA-limited step)
- `osh_transport_ion_step.c:957-960` — re-clamped again after the straggling fluctuation

So a track that's going to stop lands its exit energy on `cutoff` on the second-to-last step, and the *next* call into `ion_step_setup()` hits `e0 <= cutoff` and deletes it. This fires for **essentially every ion track that stops inside the geometry** — i.e. most of a therapy beam — not an edge case.

## Magnitude

`cutoff = max(OSH_MATERIAL_RUNTIME_EMIN, params->tcut, material_rt->emin) × A`, computed in `cutoff_total_energy()` (`osh_transport_ion_step.c:1180-1201`); `OSH_MATERIAL_RUNTIME_EMIN = 0.025` MeV/u (`src/material/runtime/osh_material_runtime.h:88`).

- **Default settings**: ≈25 keV per stopping proton track (≈0.02% of a 150 MeV beam) — below statistical noise per-history, but *systematic*, and concentrated exactly at the Bragg peak / distal edge, i.e. the region studied most closely dosimetrically.
- **Raised `TCUT0`**: scales linearly with the cutoff. `TCUT0 = 2 MeV` discards ≈1.3% of a 150 MeV proton beam's energy, entirely out of the distal dose. SH12A and FLUKA both deposit the residual energy locally at cutoff kill — this is a real behavioral gap against the reference physics, not a documented approximation.

## The fix pattern already exists in this file — for a different particle class

Sub-threshold nuclear fragments get exactly the treatment this cutoff kill is missing, via `ion_point_deposit()` (`osh_transport_ion_step.c:224-268`, added for #179):

```c
// osh_transport_ion_step.c:475-477
if (ke / (double) frag->a < OSH_TRANSPORT_ION_EMIN_MEV_PER_U) {
    /* Sub-threshold: no transportable range — point deposit here. */
    ion_point_deposit(score_rt, target, pool, slot, &ctx, species, ke, dir);
}
```

`ion_point_deposit()` builds a zero-length `struct step` at the particle's current position and hands it to `osh_scoring_score_point()` — the point-deposit API #179 added precisely for "particles that deposit their energy at a single point with no track length" (`src/scoring/runtime/osh_scoring_point.h:14-21`). A cutoff-killed ion is exactly that case. It just isn't wired to it.

## Suggested fix

Route the cutoff-kill deposit through **`osh_scoring_score_point()`**, the same way the sub-threshold-fragment path already does — i.e. reuse `ion_point_deposit()` rather than hand-rolling a second `struct step` builder.

One ordering wrinkle to work out: at the point the cutoff check runs (`ion_step_setup()`, `:557-562`), `ctx->rho` and `ctx->zone_material_idx` — both read by `ion_point_deposit()` — aren't populated yet; that happens two branches later, at `:570-571`, after the escaped-geometry check and before the blackhole check. Two reasonable options:

1. Reorder `ion_step_setup()` so zone validity + `zone_material_idx`/`rho` are resolved before the cutoff check (but still *before* the blackhole check — a particle stopping inside a blackhole/graveyard zone should stay un-scored, matching current behavior for that path).
2. Leave `ion_step_setup()`'s control flow alone and have the caller (`osh_transport_ion_step()`) issue the point deposit itself, keyed off a "killed at cutoff" reason distinct from the other early-exit reasons (unknown species, escaped geometry, blackhole) — which should *not* deposit.

Either way, the fix should touch only the cutoff-kill branch: escaped-geometry (particle legitimately leaves the scored volume carrying its energy) and blackhole (deliberate absorber, not scored by design) are working as intended and are out of scope here.

## Related, not in scope for this issue

- The same audit (finding N-1) notes the neutron cutoff kill (`osh_transport_neutron.c:302-307`, default `ncut = 1 keV`) zeroes energy with no scoring call either — though it is at least tallied into a per-run `n_cutoff` counter, unlike the ion path. The audit bundles it with a larger problem: neutron transport currently scores zero energy on *every* interaction channel, not just the cutoff kill. Worth its own issue; none filed yet.
- Two more silent-kill sites in the same function follow this pattern and are candidates for a follow-up rather than this issue:
  - Unknown species (`osh_transport_ion_step.c:551-555`) — already self-documented as `/* unknown species — kill silently */`, with no diagnostic at all, not even a warning.
  - The defensive `residual_range <= 0` kills (`:795-799`, `:904-908`) — believed unreachable given `preclip_step_len ≤ ds_csda`, but would drop the *entire* remaining energy if ever hit.
- Fragments born between `EMIN` and the run's effective cutoff (possible whenever `TCUT0 > EMIN`, since the injection threshold at `osh_transport_ion_step.c:475` checks the fixed `EMIN` constant, not `cutoff_total_energy()`) get injected only to hit this same cutoff kill on their first step. Once this issue's fix lands, that stops being a correctness problem (they'll get point-deposited too) and becomes a minor efficiency one — worth revisiting but not blocking here.

## Acceptance criteria

- [ ] An ion killed at the energy cutoff point-deposits `e0` via `osh_scoring_score_point()` at its current position, through `ion_point_deposit()` or equivalent.
- [ ] Blackhole and escaped-geometry kills continue to zero energy without scoring (unchanged behavior).
- [ ] A regression test demonstrates energy conservation for a simple stopping-in-phantom case (e.g. a low-energy proton beam in water) — total scored energy equals the primary beam energy to floating-point tolerance, at both the default cutoff and a raised `TCUT0`.

## References

- Bug-hunt audit: [P-1, 2026-07-05 deep audit](https://openshieldhit.github.io/openshieldhit/dev/bug-hunts/2026-07-05-deep-audit/#p-1) (severity/confidence: high/high; commit audited `e3c619a`, source PR #248).
- #179 — added `osh_scoring_score_point()` / the point-deposit API this issue proposes reusing.

---
<sub>Filed from the [2026-07-05 deep audit](https://openshieldhit.github.io/openshieldhit/dev/bug-hunts/2026-07-05-deep-audit/) bug-hunt report (finding P-1); the core mechanism (cutoff kill with no scoring call, and the exit-energy clamp that makes it fire on nearly every stopping track) was independently re-verified by reading `main` @ `0dbc78b` while filing this issue.</sub>

nbassler:
…ergy cutoff

### DIFF
--- a/src/transport/osh_transport_ion_step.c
+++ b/src/transport/osh_transport_ion_step.c
@@ -297,11 +297,13 @@ enum osh_status osh_transport_ion_step(struct osh_particle_pool *pool,
     /* Phase 1 — identify particle, load material, handle early exits */
     ion_step_setup(&ctx, pool, slot, zone_ref, step_segments, n_step_segments, geom_rt, transport_ctx, material_rt);
     if (ctx.done) {
-        if (ctx.stopped_at_cutoff) {
+        if (ctx.stopped_at_cutoff && ctx.rho > 0.0) {
             /* Deposit the stopping ion's residual kinetic energy at its current
              * position instead of deleting it (issue #279), attributed to the
-             * ion itself (its own generation), matching SH12A/FLUKA cutoff-kill
-             * behaviour.  Zero-length: no track-length contribution. */
+             * ion itself (its own generation).  Zero-density material has no
+             * finite CSDA residual range; a future range-to-rest terminal step
+             * for fluence would use the same positive-density precondition.
+             * Zero-length for now: no track-length contribution. */
             double stop_dir[3];
             stop_dir[0] = pool->ux[slot];
             stop_dir[1] = pool->uy[slot];
--- a/src/transport/osh_transport_ion_step.c
+++ b/src/transport/osh_transport_ion_step.c
@@ -392,7 +392,7 @@ enum osh_status osh_transport_ion_step(struct osh_particle_pool *pool,
                     np->e[k] = ev->secondaries[si].energy;
                     np->wt[k] = pool->wt[slot];
                     np->prim_idx[k] = pool->prim_idx[slot];
-                    np->gen[k] = (pool->gen[slot] < 255u) ? (uint8_t) (pool->gen[slot] + 1u) : 255u;
+                    np->gen[k] = osh_generation_child(pool->gen[slot]);
                     osh_rng_split(&np->rng[k], rng, (uint64_t) si); /* lineage+ordinal-keyed child stream */
                 } else {
                     np->n_dropped++;
@@ -414,7 +414,7 @@ enum osh_status osh_transport_ion_step(struct osh_particle_pool *pool,
             pool->e[s] = ev->secondaries[si].energy;
             pool->wt[s] = pool->wt[slot];
             pool->prim_idx[s] = pool->prim_idx[slot];
-            pool->gen[s] = (pool->gen[slot] < 255u) ? (uint8_t) (pool->gen[slot] + 1u) : 255u;
+            pool->gen[s] = osh_generation_child(pool->gen[slot]);
             pool->species[s] = ev->secondaries[si].species;
             /* Give the secondary its own stream, keyed by the parent's lineage
              * and this secondary's ordinal `si`.  Splitting reads but does not
@@ -488,7 +488,7 @@ enum osh_status osh_transport_ion_step(struct osh_particle_pool *pool,
             if (!(ke > 0.0)) {
                 continue;
             }
-            child_gen = (pool->gen[slot] < 255u) ? (uint8_t) (pool->gen[slot] + 1u) : 255u;
+            child_gen = osh_generation_child(pool->gen[slot]);
 
             if (ke / (double) frag->a < OSH_TRANSPORT_ION_EMIN_MEV_PER_U) {
                 /* Sub-threshold: no transportable range — point deposit here. */
@@ -568,7 +568,8 @@ static void ion_step_setup(struct ion_step_ctx *ctx,
 
     rc = find_projectile_index(material_rt, ctx->part, &ctx->projectile_idx);
     if (rc != OSH_OK) {
-        pool->e[slot] = 0.0; /* unknown species — kill silently */
+        pool->n_unknown_dropped++;
+        pool->e[slot] = 0.0; /* unknown species: no transport table, energy lost */
         ctx->done = 1;
         return;
     }
--- a/src/transport/osh_transport_ion_step.c
+++ b/src/transport/osh_transport_ion_step.c
@@ -220,10 +220,10 @@ static inline int _nuclear_event_elastic_scatters_primary(enum osh_nuclear_event
 /* Deposit @p energy [MeV] at the slot's current position as a zero-length point
  * (issue #179), attributed to @p species at generation @p gen.  Callers pass
  * gen = pool->gen[slot] for the slot particle's own residual (e.g. the
- * cutoff-stop deposit, #279) or gen[slot]+1 for a newly-created secondary
- * (e.g. a sub-threshold recoil fragment).  @p target is the worker-supplied
- * deposit target; NULL (or NULL fields) fall back to the shared master views
- * (issue #230). */
+ * cutoff-stop deposit, #279) or osh_generation_child(pool->gen[slot]) for a
+ * newly-created secondary (e.g. a sub-threshold recoil fragment).  @p target is
+ * the worker-supplied deposit target; NULL (or NULL fields) fall back to the
+ * shared master views (issue #230). */
 static void ion_point_deposit(struct osh_scoring_runtime *score_rt,
                               struct osh_score_target const *target,
                               struct osh_particle_pool const *pool,
--- a/src/transport/osh_transport_ion_step.c
+++ b/src/transport/osh_transport_ion_step.c
@@ -101,6 +101,7 @@ struct ion_step_ctx {
     char straggling_mode;    /* enum osh_transport_straggling_mode selected for this run */
     char is_vacuum;          /* 1 if vacuum zone or ρ ≤ 0            */
     char done;               /* 1 if step already handled (kill/nudge/error) */
+    char stopped_at_cutoff;  /* 1 if killed at the energy cutoff (deposit e0 locally, #279) */
     enum osh_status done_rc; /* return value when done == 1           */
 
     /* --- Set by ion_step_length() ---------------------------------------- */
@@ -217,10 +218,12 @@ static inline int _nuclear_event_elastic_scatters_primary(enum osh_nuclear_event
 }
 
 /* Deposit @p energy [MeV] at the slot's current position as a zero-length point
- * (issue #179), attributed to @p species and to a first-generation secondary of
- * the slot's history.  Used for the sub-threshold recoil deposit.  @p target is
- * the worker-supplied deposit target; NULL (or NULL fields) fall back to the
- * shared master views (issue #230). */
+ * (issue #179), attributed to @p species at generation @p gen.  Callers pass
+ * gen = pool->gen[slot] for the slot particle's own residual (e.g. the
+ * cutoff-stop deposit, #279) or gen[slot]+1 for a newly-created secondary
+ * (e.g. a sub-threshold recoil fragment).  @p target is the worker-supplied
+ * deposit target; NULL (or NULL fields) fall back to the shared master views
+ * (issue #230). */
 static void ion_point_deposit(struct osh_scoring_runtime *score_rt,
                               struct osh_score_target const *target,
                               struct osh_particle_pool const *pool,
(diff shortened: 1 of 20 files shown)


nbassler:
## What the fix does

Routes the energy-cutoff kill through the existing point-deposit path instead of deleting the residual kinetic energy. Five coordinated changes in `src/transport/osh_transport_ion_step.c`:

1. **Reordered `ion_step_setup()`** so the escaped-geometry, blackhole, and material-index-validity checks run *before* the cutoff check. This is what makes `ctx->zone_material_idx` / `ctx->rho` available for the deposit, and it preserves the invariant that a particle stopping in the void or a graveyard (blackhole) zone stays un-scored.
2. The cutoff branch now sets a `stopped_at_cutoff` flag (and still zeroes the pool energy / marks the step done) instead of silently dropping `e0`.
3. **The caller issues the deposit** via `ion_point_deposit()` right after setup returns `done` — matching the existing sub-threshold-fragment deposit pattern in the same function.
4. **Added a `gen` parameter to `ion_point_deposit()`.** The function previously hard-coded `gen + 1`, which is correct for a *new secondary* (a sub-threshold recoil fragment) but wrong for a stopping primary: its residual must be attributed to the ion's **own** generation (`pool->gen[slot]`), or it would be miscounted as a secondary in generation-filtered scorers.
5. Deduplicated the `gen + 1` clamp the fragment loop computed twice into one `child_gen` (which also keeps the declaration at the top of the block per DEVELOPER.md §1.2).

## Verification — energy balance

150 MeV protons into water, `NUCRE 0` (EM only, so **all** primary energy must deposit locally) and `MSCAT 0` (no lateral escape), `TCUT0 2.0` where the leak is largest:

| build | deposited MeV / primary |
|---|---|
| before fix | 148.0 — exactly 2 MeV (the cutoff) deleted per track = 1.33% of the beam |
| after fix | 150.0 — perfect energy conservation |

The 2 MeV loss matches the mechanism in the issue exactly (exit energy clamps to `cutoff` on the penultimate step, then the next setup call deletes it).

## Regression test

New `tests/unit/test_osh_cutoff_deposit.c` asserts the per-primary energy balance (`|E − 150| < 0.05 MeV`) for the setup above. Verified to **fail on the pre-fix code** (`~148 MeV`) and **pass with the fix**. Full suite: 137/137 green; clang-format clean.

Closes #279.

src/transport/osh_transport_ion_step.c
```diff
@@ -471,10 +488,11 @@ enum osh_status osh_transport_ion_step(struct osh_particle_pool *pool,
             if (!(ke > 0.0)) {
                 continue;
             }
+            child_gen = (pool->gen[slot] < 255u) ? (uint8_t) (pool->gen[slot] + 1u) : 255u;
```
grzanka: can you replace with if-ology instead of ternary operator ?
is 255 some max generation number ? maybe that should go to a well described variable and documented the max limit on generation levels.

grzanka (changes requested):

nbassler:
## Pre-merge review

Reviewed the full branch diff (all four commits) against `main`: traced every hunk plus the enclosing functions, cross-checked callers/callees, and audited against DEVELOPER.md. Branch builds clean (0 warnings), **138/138 tests pass**, clang-format clean, both new tests registered.

### No correctness bugs found — mergeable

Five things that looked risky but check out:

| Concern | Verdict |
|---|---|
| `score_neutron_point_deposit()` builds `struct step` **without** the `memset` that `ion_point_deposit()` uses | Safe — all 15 named fields of `struct step` are explicitly assigned; no uninitialized read |
| Unknown-particle test passes `NULL` for rng/score_rt/material tables | Safe — `find_projectile_index()` returns `OSH_ENOTSUP` at the `projectile_idx >= nprojectiles` bounds check *before* dereferencing `projectile_z`/`projectile_a` |
| Neutron cutoff-check moved after the `rho[mat_idx]` lookup | Safe — that lookup was already at that line for the common path; no new OOB, and escaped/blackhole still take precedence (mirrors the ion fix) |
| New `unknown_particles_dropped` counter across workers/replicas | Correct — pool-sourced and read once in `get_profile()`, exactly like `ion_secondaries_dropped`; bypasses `osh_transport_profile_merge` (which only merges timing/step counters), so no under-report |
| Docs now claim a `1e-9 MeV` neutron default + `2.53e-8 MeV` thermal clamp | Accurate — those constants (`OSH_NEUTRON_THERMAL_E_MEV`, cutoff default) pre-exist in `osh_neutron_const.h`; the docs are catching up to real behaviour |

The `osh_generation_child()` refactor is behaviour-equivalent to the old `(gen < 255) ? gen+1 : 255` ternary at all five call sites.

### DEVELOPER.md — compliant

- **§6.1** (`static inline` naming): `osh_generation_child` uses the `osh_` prefix, not a leading `_`. Correct — the leading-underscore rule is for *file-local private* static inline helpers; this is a **shared** header utility, matching the precedent (`osh_physics_beta`, `osh_voxel_tile_idx`, `osh_rng`).
- **§1.1 / §1.2** (declaration placement): `stop_dir` and `child_gen` are declared at the top of their blocks and assigned late — consistent with the rule and surrounding code.

### One minor note (not a blocker)

The neutron cutoff-deposit guards `if (rho > 0.0)` before depositing; the **ion** cutoff-deposit does not. It is **not** a divide-by-zero — the DOSE point estimator has its own `if (!(st->rho > 0.0)) return OSH_OK;` guard (`osh_scoring_estimator_point.c:71`) — and a stopping ion in a zero-density zone is essentially unreachable (ions do not lose energy in vacuum, so they never arrive at cutoff there). The only asymmetry: *if* it ever happened, the ion path would book `Energy` (but not `Dose`) at that vacuum location, whereas the neutron path books nothing. Harmless, but adding `&& ctx.rho > 0.0` to the `stopped_at_cutoff` deposit condition would make the two paths visibly symmetric. Purely optional.

### Recommendation: merge

The core #279 fix is verified by energy balance (148 → 150 MeV/primary); the two related fixes (neutron residual, unknown-particle accounting) are clean and consistently follow the ion pattern; everything is tested and green.

nbassler:
This PR helps a lot for the missing dose in the distal end of SOBPs from the https://github.com/APTG/2022_DCPT_LET repository

image

src/transport/osh_transport_neutron.c
```diff
@@ -128,6 +129,52 @@ static enum osh_status score_neutron_step(struct osh_scoring_runtime *score_rt,
...
+    st.de = energy;
+    st.rho = rho;
+    st.wt = pool->wt[k];
+    st.voxel_idx = (zone_ref && zone_ref->has_hu) ? zone_ref->voxel_idx : 0u;
+    st.medium = zone_ref ? (int) zone_ref->material_idx : -1;
+    st.zone = zone_ref ? (int) zone_ref->zone_idx : -1;
+    st.system = OSH_COORD_UNIVERSE;
+    st.prim_idx = pool->prim_idx[k];
+    st.gen = pool->gen[k];
+    st.has_voxel = (zone_ref && zone_ref->has_hu) ? 1u : 0u;
```
grzanka: @nbassler can you get rid of ternaries replacing them with proper if statements with comments ?

grzanka (changes requested):
fix ternary stuff

grzanka:
@nbassler could you also update `docs/dev/bug-hunts/2026-07-05-deep-audit.md` as part of this PR (or a quick follow-up)?

The **P-1** finding in that report ("Residual kinetic energy is silently deleted when a particle is killed at the energy cutoff") is exactly what this PR fixes — it even matches the `TCUT0 = 2 MeV` → ~1.3% beam-energy-loss figure used in the new regression test. Right now the doc still describes it as an open finding, and the "Follow-up" line at the top only records `#255→#257` and `#267→a0f60a1` as filed/fixed. If this merges as-is, the bug-hunt report will read as though the leak is still present.

Could you:
- add `#279` / this PR to the "Follow-up" summary line at the top of the doc, and
- add a "fixed by #285" note under the P-1 entry itself (matching the style already used for the other two resolved findings)?

That P-1 entry also flags the related "unknown species killed with no diagnostic" issue, which this PR fixes too — worth covering both in the same update.

---

grzanka (approved):

PR #286, opened 2026-07-08
feat: Enhance differential binning and units for scoring outputs and …

### LINKED ISSUE #215: SCORING: differential (Diff1/Diff2) output is counts-per-bin, not a spectral density (missing bin-width normalisation)
## Summary

Differential scoring (`Diff1`/`Diff1Type`, and `Diff2`) accumulates the raw
quantity **summed into each differential bin** — i.e. Φ(bin) — and never divides
by the differential **bin width**. The output is therefore counts-per-bin, not a
spectral **density** (e.g. dΦ/dE_kin). This is inconsistent with SHIELD-HIT12A,
which reports the same estimator already normalised per unit axis, and it is
especially misleading for **logarithmic** binning, where bin widths span orders
of magnitude.

Concretely, for a `Quantity Fluence` scorer with `Diff1 0.1 300 150 LOG` /
`Diff1Type EKIN`:

- **SHIELD-HIT12A** writes each page `[/cm^2/MeV]` (already dΦ/dE_kin).
- **OpenShieldHIT** writes `[/cm^2]` per bin (Φ per bin). Neighbouring log bins
  differ in width by `exp(ln(hi/lo)/nbins)`, so the OSH curve has a systematic
  bin-width tilt that SH12A's does not — the two are not directly comparable
  without post-hoc division.

## Where

- No normalisation in `src/scoring/runtime/osh_scoring_postprocess.c` (only the
  per-primary NORM division happens; there is no `diff_nbins > 0` branch).
- The differential axis metadata needed to normalise is already present on the
  page (`diff_lo/diff_hi/diff_nbins/diff_log`, `src/scoring/runtime/osh_scoring_output_runtime.h`)
  and is written to both output formats
  (`src/scoring/save/osh_scoring_save_ascii.c`, `..._bdo2019.c`), so a downstream
  tool *can* reconstruct the widths — but the native output is not a density.

## Impact

- `tools/plot_nucre.py` (issue #212 NUCRE decks) currently works around this by
  dividing the OSH differential by the reconstructed log-bin width before
  overlaying against the SH12A `spectrum.dat` (which is already a density).
- Same asymmetry affects any comparison of the `NB_target_diff` /
  `NB_target_water_diff` differential fluence pages in
  `tests/cases/10_plan01_geoA_sobpcent` / `11_plan02_geoD_mono` against SH12A.
- `tests/cases/09_diff_scoring` exercises the feature but does not check the
  normalisation.

## Proposed fix

Divide each differential bin by its axis bin width during postprocessing for
`diff_nbins > 0` pages (and `diff2_nbins > 0` for the second axis), handling both
linear and logarithmic binning:

- linear: `Δ = (hi - lo) / nbins`
- log:    `Δ_b = lo * (ratio^((b+1)/nbins) - ratio^(b/nbins))`, `ratio = hi/lo`

Then update the unit labels in the ASCII/BDO save paths to include the per-axis
unit (e.g. `[/cm^2/MeV]` for a Fluence-vs-EKIN page), so the output is a true
spectral density matching SH12A.

## Caveats

- This is a **behaviour change** for the differential output: any existing
  consumer that assumes counts-per-bin (including the current `plot_nucre.py`
  workaround) must be updated to stop dividing.
- Decide the convention for LET/QEFF differential axes and for double-differential
  pages (divide by the product of both bin widths).
- Consider whether `Diff1Type` with a stopping-power (`DEDX`/`LET`) axis should
  normalise by the LET-bin width in the same way.

Found while building the issue #212 NUCRE reference decks.

### LINKED ISSUE #282: BDO2019: clarify LOG differential binning and units for EKIN spectra
## Summary

When scoring target kinetic-energy fluence spectra with logarithmic binning, the generated BDO should preserve the LOG binning metadata consistently and expose clear differential units.

Example input:

```text
Output
        Filename NB_target_diff.bdo
        Geo TARGET

        Quantity Fluence
        Diff1     0.1  300.0  150  LOG
        Diff1Type EKIN
```

## Observed

In an OpenShieldHit-generated `NB_target_diff.bdo`, the EKIN pages contain:

```text
PAG_DIF_SET   = [-1]
PAG_DIF_TYPE  = [1, 0]
PAG_DIF_START = [0.1, 0.0]
PAG_DIF_STOP  = [300.0, 1.0]
PAG_DIF_SIZE  = [150, 1]
PAG_DIF_UNITS = ["MeV;"]
PAG_DATA_UNIT = ["1/cm^2"]
```

So the LOG hint seems to be present via `PAG_DIF_SET = -1`.

However, the unit metadata is ambiguous for downstream tools. The quantity is a differential fluence spectrum, but `PAG_DATA_UNIT` is non-differential and `PAG_DIF_UNITS` only lists the differential axis unit. In comparable SHIELD-HIT12A output, the corresponding fields look like:

```text
PAG_DIF_UNITS = ["/cm^2/MeV"]
PAG_DATA_UNIT = ["/cm^2;MeV"]
```

## Expected

For a `Diff1 ... LOG` spectrum, the BDO should contain enough metadata for downstream readers to know:

1. The differential axis is logarithmically binned.
2. The bin range is `[0.1, 300.0]`.
3. The bin count is `150`.
4. The plotted/scored quantity is differential fluence, i.e. effectively `1/(cm^2 primary MeV)` after normalization.

## Why this matters

Downstream conversion via `convertmc plotdata` currently emits linear-looking x centers for this spectrum:

```text
1.09967, 3.099, 5.09833, ...
```

For `0.1 300 150 LOG`, expected log-bin centers start near:

```text
0.1027, 0.1083, ...
```

This may be partly a `convertmc` issue, but OpenShieldHit should make the BDO metadata unambiguous and consistent.

## Request

Please confirm whether `PAG_DIF_SET = -1` is the intended LOG marker in BDO2019 and whether the unit metadata for differential fluence spectra should be adjusted to explicitly encode `1/cm^2/MeV`-style differential units.

nbassler:
…documentation

### DIFF
--- a/docs/dev/bug-hunts/2026-07-05-deep-audit.md
+++ b/docs/dev/bug-hunts/2026-07-05-deep-audit.md
@@ -427,16 +427,15 @@ exactly at lung/soft-tissue/bone interfaces where dose accuracy is most
 scrutinised. (Also note the step is scored on the chord `p→q` while the
 transport step may have been hinge-bent — documented O(θ₀²·L), fine.)
 
-### S-4 (low, high) Differential-axis edge semantics: both edges exclusive, out-of-range silently dropped, ASCII spectra not divided by bin width {: #s-4 }
+### S-4 (low, high) Differential-axis edge semantics: both edges exclusive, out-of-range silently dropped, ASCII spectra now divided by bin width {: #s-4 }
 
 `diff_axis_bin` (`osh_scoring_step.c:407-421`) treats the axis as the *open*
 interval `(lo, hi)`: a value exactly at `lo` (or `hi`) is discarded via the
 out-of-range sentinel, and there are no under/overflow bins anywhere, so the
-discarded tally is invisible. Combined with the ASCII writer normalising only
-by `1/nstat` (`osh_scoring_save_ascii.c:333-340`) and *not* by bin width, a
-log-binned "spectrum" output is a per-bin count whose shape differs from the
-physical dN/dE — fine as a convention, but it is not documented in the
-writer header, which advertises "immediately human-readable".
+discarded tally is invisible. The original audit also found that differential
+outputs were written as per-bin counts; issue #215 moved bin-width
+normalisation into scoring postprocess, so ASCII and BDO2019 writers now receive
+differential quantities.
 
 ### S-5 (info, high) Verified-OK: the accumulator/merge layer is sound and correctly shaped for #169 {: #s-5 }
 
--- a/docs/dev/scoring.md
+++ b/docs/dev/scoring.md
@@ -300,13 +300,15 @@ Each estimator owns a trio of handlers, registered in
 
 A `—` below means the handler is `NULL`: the estimator is not scored on that path
 (`score_point_` needs a track length for FLUENCE/LET), or the accumulator is
-already final (`postprocess_` for ENERGY/COUNT). The scorer deposits the extensive
-quantity; the *only* geometry-specific normalisation, ÷volume, lives in
-`postprocess_` (via `geo->bin_vol_inv`) — never at score time and never at save.
+already final (`postprocess_` for ENERGY/COUNT unless the page is differential).
+The scorer deposits the extensive quantity; geometry-specific normalisation
+(÷volume via `geo->bin_vol_inv`) and differential-axis normalisation (÷bin width,
+or ÷width1÷width2 for double-differential pages) live in `postprocess_` — never
+at score time and never at save.
 
 | `Quantity` (score kind) | `score_step_` | `score_point_` | `postprocess_` | deposits → finalises |
 |---|---|---|---|---|
-| `ENERGY`  | `osh_scoring_estimator_step_energy`  | `osh_scoring_estimator_point_energy` | — | `de·(path/score_len)` → already final [MeV] |
+| `ENERGY`  | `osh_scoring_estimator_step_energy`  | `osh_scoring_estimator_point_energy` | differential pages only | `de·(path/score_len)` → [MeV] or [MeV/axis-unit] |
 | `FLUENCE` | `osh_scoring_estimator_step_fluence` | — | `postprocess_volume` | track length → ÷volume [1/cm²] |
 | `DOSE`    | `osh_scoring_estimator_step_dose`    | `osh_scoring_estimator_point_dose`  | `postprocess_volume` | `de·(path/score_len)/ρ` [+SP-ratio] → ÷volume [MeV/g] |
 | `DOSEGY`  | `osh_scoring_estimator_step_dose`    | `osh_scoring_estimator_point_dose`  | `postprocess_dosegy` | as `DOSE` → ÷volume, ×`OSH_MEVG2GY` [Gy] |
--- a/docs/user/detect.dat.md
+++ b/docs/user/detect.dat.md
@@ -147,8 +147,9 @@ Additive differential pages are reported as differential quantities: after
 spatial postprocessing, each bin is divided by the width of its `Diff1` bin, and
 by the product of `Diff1` and `Diff2` widths for double-differential pages.  For
 example, `Quantity Fluence` with `Diff1Type EKIN` is reported per MeV
-(`1/cm²/MeV`), not as fluence accumulated per energy bin.  `LOG` binning uses
-the actual logarithmic bin widths for this normalisation.
+(`/cm²/MeV`, equivalent to `1/cm²/MeV`), not as fluence accumulated per energy
+bin.  `LOG` binning uses the actual logarithmic bin widths for this
+normalisation.
 
 ### Axis stopping-power override
 
--- a/docs/user/detect.dat.md
+++ b/docs/user/detect.dat.md
@@ -143,6 +143,12 @@ Supported axis type keywords (same for Diff1Type and Diff2Type):
 
 Differential scoring is supported for `Fluence`, `Dose`, `DoseGy`, and `Energy`.
 Averaged quantities (`DLET`, `TLET`, `DQEFF`, `TQEFF`) cannot carry a differential axis.
+Additive differential pages are reported as differential quantities: after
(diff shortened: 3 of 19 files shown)


nbassler:
## Design review

Reviewed the branch (postprocess + save + shadow changes for #215/#282). Branch builds clean (0 warnings); all scoring tests pass.

### Verdict: the design is right — putting the dy/dx conversion in postprocess is the correct layer.

Three reasons it beats the alternatives:

1. **It composes with the normalizations already there.** Per-primary NORM and per-volume division already live in postprocess and turn extensive accumulators into intensive quantities; bin-width division is the same kind of operation. Confirmed FLUENCE goes `raw × vinv × 1/nstat × 1/Δbin` in one pass (`page_scale_by_bin_volume`), i.e. exactly `dΦ/dE`. Doing it at save time would duplicate the logic across the ASCII and BDO writers and desync from the volume/NORM math.

2. **Raw accumulators survive, so checkpointing stays correct.** Because the live accumulators stay as per-bin sums, every incremental (shadow) save re-derives the density from raw — no drift, no double-division. The critical enabler is extending `osh_scoring_postprocess_page_writes_data()` so the shadow allocates **scratch** for a differential `Energy` page; without it, the out-of-place transform would alias and corrupt the live accumulator on the first checkpoint. That's the single most important correctness point in the PR and it's handled correctly.

3. **The save formats are genuinely agnostic** — they serialize whatever postprocess produced; the BDO just relabels units. Clean separation.

Also verified: the two dispatch paths are mutually exclusive (FLUENCE via the estimator handler, Energy via the `SUM/NORM` switch), so **no double-division**; the log-width formula matches #215's spec exactly; and excluding `AVER` is correct (dividing a dose-averaged LET by ΔE would be meaningless).

### Suggested changes (none are blockers)

- **Add a direct out-of-place test for the checkpoint path.** The shadow / differential-Energy path — the highest-risk scenario — is only covered *indirectly* (the predicate test proves scratch *would* be allocated; the normalization tests are all in-place). Add one test that runs `osh_scoring_postprocess_into(dst, src)` with a distinct `dst` on a differential Energy page and asserts (a) `dst` scratch holds the density and (b) `src` live accumulator is **byte-unchanged**, and ideally that two successive snapshots give the same density (no accumulator drift). That directly locks in the property the design depends on.

- **The two BDO unit helpers are confusingly named and near-duplicate.** `page_data_unit_with_diff` (uses `;`) vs `page_differential_unit_with_diff` (uses `/`) differ only in the separator, and their names don't hint at which BDO token they feed. Either collapse into one helper taking a separator char, or rename to say what they produce (`PAG_DATA_UNIT = "data;axis"` vs `PAG_DIF_UNITS = "data/axis"`), with a one-line comment mapping each to its token.

- **`1/cm^2` vs SH12A's `/cm^2`.** Output is `1/cm^2;MeV` / `1/cm^2/MeV`; #282 shows SH12A writes `/cm^2;MeV` / `/cm^2/MeV`. The leading `1` is harmless for a human but differs from the reference format being synced to — decide whether exact SH12A-parser compatibility matters and strip it if so.

- **Comment the *why* of the `AVER`/`APPEND` exclusion** in `page_needs_diff_normalization()`. It's correct, but a future dev "fixing" it to also normalize DLET would introduce a subtle physics bug. One line ("averages and appended raw values are not densities") prevents that.

- **Settle one design point before the future merge lands:** with density in the BDO, combining multiple runs' BDO files must merge by **nstat-weighted average**, not by summing (summing densities is wrong). That works for per-primary NORM quantities and matches SH12A, but it's a constraint the merge tool must honor — worth noting in the postprocess header comment next to the `#170` note so the merge author doesn't assume additive BDOs.

Net: the architecture is sound; I wouldn't restructure any of it. The above is polish plus one meaningful test-coverage gap on the checkpoint path.

nbassler:
> ### Suggested changes (none are blockers)
> * **Add a direct out-of-place test for the checkpoint path.** The shadow / differential-Energy path — the highest-risk scenario — is only covered _indirectly_ (the predicate test proves scratch _would_ be allocated; the normalization tests are all in-place). Add one test that runs `osh_scoring_postprocess_into(dst, src)` with a distinct `dst` on a differential Energy page and asserts (a) `dst` scratch holds the density and (b) `src` live accumulator is **byte-unchanged**, and ideally that two successive snapshots give the same density (no accumulator drift). That directly locks in the property the design depends on.

resloved.

> * **The two BDO unit helpers are confusingly named and near-duplicate.** `page_data_unit_with_diff` (uses `;`) vs `page_differential_unit_with_diff` (uses `/`) differ only in the separator, and their names don't hint at which BDO token they feed. Either collapse into one helper taking a separator char, or rename to say what they produce (`PAG_DATA_UNIT = "data;axis"` vs `PAG_DIF_UNITS = "data/axis"`), with a one-line comment mapping each to its token.

resloved.

> * **`1/cm^2` vs SH12A's `/cm^2`.** Output is `1/cm^2;MeV` / `1/cm^2/MeV`; [BDO2019: clarify LOG differential binning and units for EKIN spectra #282](https://github.com/openshieldhit/openshieldhit/issues/282) shows SH12A writes `/cm^2;MeV` / `/cm^2/MeV`. The leading `1` is harmless for a human but differs from the reference format being synced to — decide whether exact SH12A-parser compatibility matters and strip it if so.

resloved. It is a bug in SH12A which we should not carry over into OSH.

> * **Comment the _why_ of the `AVER`/`APPEND` exclusion** in `page_needs_diff_normalization()`. It's correct, but a future dev "fixing" it to also normalize DLET would introduce a subtle physics bug. One line ("averages and appended raw values are not densities") prevents that.

resloved.

> * **Settle one design point before the future merge lands:** with density in the BDO, combining multiple runs' BDO files must merge by **nstat-weighted average**, not by summing (summing densities is wrong). That works for per-primary NORM quantities and matches SH12A, but it's a constraint the merge tool must honor — worth noting in the postprocess header comment next to the `#170` note so the merge author doesn't assume additive BDOs.

resloved by clarifying how it is handled.

PR #287, opened 2026-07-08
feat: Update Bohr variance calculations to include relativistic facto…

### LINKED ISSUE #280: physics: Bohr Gaussian straggling omits the relativistic variance factor (σ ~11% low at 200 MeV; discontinuity at STRAGG 2's κ=10 dispatch)
## Summary

`osh_physics_strag_sigma()` — the Bohr Gaussian energy-straggling width (`src/physics/atomic/osh_physics_strag_gauss.c:24-36`) — implements only the **classical, non-relativistic** Bohr variance:

```
σ² = C_bohr · z_eff² · (Z/A) · d      (no β or γ dependence)
```

The physically correct thick-absorber limit (PDG "Passage of particles through matter"; equivalently, the Gaussian limit of the Vavilov distribution this codebase already implements) is `σ² = ξ·E_max·(1−β²/2)`, which — using this module's own `ξ` and `E_max` — works out to exactly `σ²_classical × γ²(1−β²/2)`. The missing factor grows with energy:

| proton energy | γ²(1−β²/2) | σ error |
|---|---|---|
| 70 MeV | 1.078 | −3.8% |
| 200 MeV | 1.236 | −10.5% |
| 1 GeV | ~1.9 | −27% |

## Where this bites

`osh_physics_strag_sigma()` is called from exactly two places, both in `osh_transport_ion_step.c`, and **neither passes β²**:

1. **STRAGG 1 (pure Gaussian)** — the call at `osh_transport_ion_step.c:952`, inside the non-Vavilov branch (`:951-954`) of the straggling block. `gamma`/`beta2` aren't even computed on this code path — they're local to the `if (VAVILOV)` branch above. Every Gaussian-mode run understates straggling width at all energies, worst at high energy: the distal falloff comes out systematically too sharp.
2. **STRAGG 2 (auto Vavilov/Landau/Gaussian by κ)** — the call at `:942`, inside the `κ ≥ OSH_STRAG_KAPPA_GAUSS` branch of the dispatcher (`:941-943`; `OSH_STRAG_KAPPA_GAUSS = 10.0`, `src/physics/atomic/osh_physics_strag.h:38`). Here `beta2`/`gamma` **are already computed two lines above** (`:936-937`) for the ξ/E_max/κ calculation — but the call at `:942` drops them on the floor. Since the Vavilov/Landau branch (`κ < 10`) samples from a distribution whose variance already carries the correct `ξ·E_max(1−β²/2)` scaling, the model's variance **discontinuously jumps by γ²(1−β²/2) exactly at the κ=10 boundary** — a physics artifact as a function of step thickness/energy, not just a magnitude error.

## Verification

I re-derived the missing factor from this module's own formulas rather than taking the audit's number on faith:

- `osh_physics_strag_xi()`: `ξ = (K/2)·(Z/A)·(z_eff²/β²)·d` (`src/physics/atomic/osh_physics_strag.c:7-13`)
- `osh_physics_strag_emax()`: `E_max ≈ 2·mₑc²·β²γ²` for `mₑ≪M` (`osh_physics_strag.c:15-28`)
- So `ξ·E_max = K·mₑc²·(Z/A)·z_eff²·d·γ² = σ²_classical·γ²` (since `C_bohr = K_bethe × m_e c²`, per the comment at `osh_physics_strag_gauss.c:6-8`), and the Gaussian-limit variance is `ξ·E_max·(1−β²/2) = σ²_classical·γ²(1−β²/2)`.
- Plugging in this repo's own existing test fixture values for a 200 MeV proton in water (`tests/unit/test_osh_physics_strag.c`: `β² = 0.320525`, `ξ = 0.026590` MeV, `E_max = 0.4815` MeV, `z_over_a = 0.5551`) reproduces the table above to within test-fixture rounding: `σ_classical ≈ 0.0933` MeV vs. `σ_relativistic ≈ 0.1037` MeV, a ratio of 1.111 (σ ≈ 10% low), matching the audit's −10.5% at 200 MeV.

## Why existing validation didn't catch it

Per the audit, the STRAGG 2 SH12A distal-falloff comparison in `TODO.md` mostly exercises the regime where the Vavilov/Landau branch is taken (`κ < 10`, i.e. `γ → 1`), where the missing factor is small. Separately (audit finding T-2), the `reference::idd_*` comparison tier that would actually exercise distal-falloff width doesn't run in CI at all today (no reference curves committed) — so nothing is asserting on this regardless. The discontinuity is also reachable at high energy with **thick steps** (coarse `DELTAE` in a thick slab pushes `κ` past 10 even for a relativistic particle), a case no existing test covers.

## Suggested fix

Thread `beta2` into `osh_physics_strag_sigma()` itself — mirroring `osh_physics_strag_xi(z_eff, z_over_a, thickness_gcm2, beta2)`, which already takes it — and apply the correction once, inside the shared function, rather than duplicating it at both call sites:

- **The function returns σ, not σ².** The `γ²(1−β²/2)` factor applies to the *variance*, so the returned `σ` must be scaled by `sqrt(γ²(1−β²/2)) = γ·√(1−β²/2)`, not by `γ²(1−β²/2)` directly. Easy to get backwards — worth an inline comment given this file's existing habit of spelling out formulas.
- At the STRAGG 2 call site (`:942`), `beta2` is already in scope — pass it straight through.
- At the STRAGG 1 call site (`:952`), compute `gamma`/`beta2` from `e_mid`/`ctx->proj_mass_mev` the same way the Vavilov branch already does (`:936-937`), then pass it.
- This also removes the κ=10 discontinuity for free, since both sides of the dispatch then use the same relativistic variance.

Centralizing the correction in `osh_physics_strag_gauss.c` (rather than scaling `sigma_strag` after the call at each of the two sites in the transport hot path) keeps it consistent with how `osh_physics_strag.c` already centralizes ξ/E_max/κ for the per-model samplers to share.

## Test impact

- `tests/unit/test_osh_physics_strag.c:54-61` (`test_sigma_guards_and_scaling`) calls the current 3-argument signature directly and needs updating for the new parameter.
- Worth adding, in the same hand-derived style already used in that file (see its header comment: reference values "hand-computed from the analytic definitions... not copied from any tabulated third-party numbers"): a case asserting the σ ratio with/without the relativistic factor matches `γ²(1−β²/2)` at a known energy (e.g. the file's existing 200 MeV proton / water fixture, factor ≈1.236), and a continuity check that the Gaussian-branch and Vavilov-branch variances agree at `κ` on either side of `OSH_STRAG_KAPPA_GAUSS`.

## Acceptance criteria

- [ ] `osh_physics_strag_sigma()` applies the `γ²(1−β²/2)` relativistic correction to the Bohr variance (both STRAGG 1 and the STRAGG 2 Gaussian branch).
- [ ] No discontinuity at the STRAGG 2 dispatch boundary: σ from the Gaussian branch and σ implied by the Vavilov/Landau branch agree in the `κ → 10` limit.
- [ ] Regression test locks in the corrected σ at a reference energy (e.g. 200 MeV proton in water) against a hand-derived value, in the style already used in `test_osh_physics_strag.c`.
- [ ] `tests/unit/test_osh_physics_strag.c` updated for the new signature/behavior.

## Related, not in scope for this issue

- The audit's T-2 finding (the `reference::idd_*` physics-validation tier doesn't run in CI; no reference curves committed) is a separate, already-documented infrastructure gap — not fixed here, but it's part of why this was never caught.

## References

- Bug-hunt audit: [P-3, 2026-07-05 deep audit](https://openshieldhit.github.io/openshieldhit/dev/bug-hunts/2026-07-05-deep-audit/#p-3) (severity/confidence: medium/high; commit audited `e3c619a`, source PR #248).
- #208 — added the Vavilov/Landau straggling module this bug lives in.

---
<sub>Filed from the [2026-07-05 deep audit](https://openshieldhit.github.io/openshieldhit/dev/bug-hunts/2026-07-05-deep-audit/) bug-hunt report (finding P-3). The missing relativistic factor and the κ=10 discontinuity were independently re-derived from the module's own ξ/E_max formulas and cross-checked against this repo's own test fixture values while filing this issue (see "Verification" above), not just re-stated from the audit. Line numbers verified against `main` @ `0dbc78b`.</sub>

nbassler:
…rs and enhance unit tests

### DIFF
--- a/src/physics/atomic/osh_physics_strag.c
+++ b/src/physics/atomic/osh_physics_strag.c
@@ -4,6 +4,14 @@
 
 #include "openshieldhit/const.h"
 
+static double emax_from_gamma_beta2(double gamma, double mass_mev, double beta2) {
+    double ratio; /* electron-to-projectile mass ratio mₑ/M       */
+
+    ratio = OSH_ELECTRON_MASS_MEV / mass_mev;
+    /* E_max = 2·mₑc²·β²γ² / (1 + 2γ·mₑ/M + (mₑ/M)²) */
+    return 2.0 * OSH_ELECTRON_MASS_MEV * beta2 * gamma * gamma / (1.0 + 2.0 * gamma * ratio + ratio * ratio);
+}
+
 double osh_physics_strag_xi(double z_eff, double z_over_a, double thickness_gcm2, double beta2) {
     if (z_eff <= 0.0 || z_over_a <= 0.0 || thickness_gcm2 <= 0.0 || beta2 <= 0.0) {
         return 0.0;
@@ -21,20 +29,17 @@ double osh_physics_strag_emax(double t_kin_mev, double mass_mev) {
     }
     gamma = 1.0 + t_kin_mev / mass_mev;
     beta2 = 1.0 - 1.0 / (gamma * gamma);
-    return osh_physics_strag_emax_beta2(t_kin_mev, mass_mev, beta2);
+    return emax_from_gamma_beta2(gamma, mass_mev, beta2);
 }
 
 double osh_physics_strag_emax_beta2(double t_kin_mev, double mass_mev, double beta2) {
     double gamma; /* Lorentz factor γ = 1 + T/M                    */
-    double ratio; /* electron-to-projectile mass ratio mₑ/M       */
 
     if (t_kin_mev <= 0.0 || mass_mev <= 0.0 || beta2 <= 0.0 || beta2 >= 1.0) {
         return 0.0;
     }
     gamma = 1.0 + t_kin_mev / mass_mev;
-    ratio = OSH_ELECTRON_MASS_MEV / mass_mev;
-    /* E_max = 2·mₑc²·β²γ² / (1 + 2γ·mₑ/M + (mₑ/M)²) */
-    return 2.0 * OSH_ELECTRON_MASS_MEV * beta2 * gamma * gamma / (1.0 + 2.0 * gamma * ratio + ratio * ratio);
+    return emax_from_gamma_beta2(gamma, mass_mev, beta2);
 }
 
 double osh_physics_strag_kappa(double xi, double e_max) {
--- a/src/physics/atomic/osh_physics_strag.c
+++ b/src/physics/atomic/osh_physics_strag.c
@@ -15,13 +15,23 @@ double osh_physics_strag_xi(double z_eff, double z_over_a, double thickness_gcm2
 double osh_physics_strag_emax(double t_kin_mev, double mass_mev) {
     double gamma; /* Lorentz factor γ = 1 + T/M                    */
     double beta2; /* projectile β² = 1 − 1/γ²                      */
-    double ratio; /* electron-to-projectile mass ratio mₑ/M       */
 
     if (t_kin_mev <= 0.0 || mass_mev <= 0.0) {
         return 0.0;
     }
     gamma = 1.0 + t_kin_mev / mass_mev;
     beta2 = 1.0 - 1.0 / (gamma * gamma);
+    return osh_physics_strag_emax_beta2(t_kin_mev, mass_mev, beta2);
+}
+
+double osh_physics_strag_emax_beta2(double t_kin_mev, double mass_mev, double beta2) {
+    double gamma; /* Lorentz factor γ = 1 + T/M                    */
+    double ratio; /* electron-to-projectile mass ratio mₑ/M       */
+
+    if (t_kin_mev <= 0.0 || mass_mev <= 0.0 || beta2 <= 0.0 || beta2 >= 1.0) {
+        return 0.0;
+    }
+    gamma = 1.0 + t_kin_mev / mass_mev;
     ratio = OSH_ELECTRON_MASS_MEV / mass_mev;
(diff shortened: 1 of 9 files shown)


nbassler:
[straggling_report_new.pdf](https://github.com/user-attachments/files/29816666/straggling_report_new.pdf)

nbassler:
image

nbassler:
Benchmark spot-check, nstat=20000:
```
case    OSH before   OSH after r1   OSH after r2
strag0     6.04 s        6.04 s        6.03 s
strag1     7.36 s        7.35 s        7.37 s
strag2     8.41 s        8.32 s        8.30 s
```
The effect is where expected: STRAGG 2 improves by roughly 1%, while
STRAGG 0/1 are within run-to-run timing noise. SH12A-relative speedups in
the after runs were 2.35x and 2.39x for STRAGG 2.

nbassler:
[straggling_report_r2.pdf](https://github.com/user-attachments/files/29817224/straggling_report_r2.pdf)

nbassler:
Repeated local benchmark spot-checks show STRAGG 2 improving from about
8.41 s to 8.30-8.32 s at nstat=20000. STRAGG 0/1 stayed within timing noise,
as expected because this patch targets the Vavilov-mode dispatch path.

[straggling_report_r3.pdf](https://github.com/user-attachments/files/29817456/straggling_report_r3.pdf)

PR #319, opened 2026-07-28
random: fold RNDOFFSET into the run seed instead of the history index

### LINKED ISSUE #317: -N/--seedoffset is a history-index offset, not a stream selector — parallel array jobs correlate
## Summary
`-N`/`--seedoffset` does not behave like SHIELD-HIT12A's seed offset. In SH12A `-N`
selects an independent RNG stream, so `generatemc`/`mcpartools` decorrelates array
jobs simply by handing each a distinct small `-N` (1..100). In OSH `-N` sets a
**history-index base** (`hist_base`), and job independence instead requires
**disjoint index ranges**. With the generatemc convention (consecutive small
offsets) and a large per-job `nstat`, the index ranges overlap almost completely,
so the jobs are near-duplicates rather than independent replicas.

This surfaced while adapting the SH12A `generatemc` HPC workflow to OpenShieldHIT.

## Evidence (local build, plan02 mono, `nstat=2000`)
`-N` is honored and reproducible:
- `-N 0` -> dose `0.05353392204249136`; `-N 5000` -> `0.06124823919745243`; `-N 0` again -> identical.

But consecutive offsets overlap:

```
-N0    baseline            : 0.05353392204249136
-N1    overlap 1999/2000   : 0.05353392204249136   -> |d| = 0.0000%   (near-duplicate)
-N2000 disjoint            : 0.06048367728613643   -> |d| = 12.98%    (independent)
```

`-N 0` and `-N 1` are effectively the same run (they share 1999/2000 histories).
So `generatemc` giving jobs `-N = 1..100` with `nstat = 1e6` would produce ~100
correlated near-duplicates -- biased statistics with a badly underestimated variance.

## Root cause
- `src/transport/osh_transport_ion.c:145-146` -- `seeding.seed = rndseed;
  seeding.hist_base = rndoffset;` so `-N` shifts the index, not the stream.
- `src/random/osh_rng.c:83` `osh_rng_seed_history()` derives the stream as
  `f(seed, hist_index, purpose)`; two runs sharing a `hist_index` share the stream.
- `hist_base`/index-offset semantics are documented as the splitting mechanism in
  `src/random/osh_rng.h` and the `random`/`transport` READMEs, and the field doc in
  `src/transport/osh_transport.h` explicitly states RNDOFFSET is "Not a value added
  to rndseed."

## Proposed direction (for discussion)
Make `-N` a **stream selector** (SH12A-style): fold it into the run seed instead of
the index --
`seeding.seed = rndseed + rndoffset; seeding.hist_base = 0;`
Because `seed` drives both the PCG init seed and the mixed stream constant, distinct
`-N` then give fully independent, reproducible stream families over the same
`[0, nstat)` range. The 9999 cap stays fine (100 << 9999), and no-`-N` runs are
byte-identical (seed = rndseed + 0), so existing goldens are unaffected.

Test review: `test_osh_beam_runtime_fill` (sets `hist_base` directly) and
`test_osh_cli` (parsing + 9999 cap) and the `test_osh_simulation` invariance test
all remain valid under this change; case goldens run without `-N`.

## Open questions
1. Keep `hist_base` as a *separate* knob for intra-run range splitting (MPI/process),
   decoupled from `-N`? (The `osh_beam_runtime` plumbing + `test_osh_beam_runtime_fill`
   would stay; only the `-N -> hist_base` wiring is removed.)
2. Additive `rndseed + rndoffset`, or a hashed combine to avoid `(seed, -N)` sum-aliasing?
3. Keep the flag name / `RNDOFFSET` card, or introduce an explicit `--rndseed` / `RNDSEED`
   override alongside?

grzanka:
## Summary

Fixes #317.

`-N`/`--seedoffset` (`RNDOFFSET`) shifted the **history-index base** rather than selecting an **independent RNG stream**. Confirmed locally (`plan02_geoD_mono`, `nstat=2000`): `-N 0` vs `-N 1` produced **byte-identical** scored `.bdo` payloads (the only diff was the embedded output-path metadata), while `-N 2000` (a disjoint range) differed throughout — because `-N 0`/`-N 1` share 1999/2000 of their history-index range, and the RNG stream only depends on `(rndseed, hist_index)`, never on `rndoffset`.

This breaks the documented "embarrassingly-parallel / multi-node" BDO merge path (`docs/dev/scoring.md`) for exactly the HPC array-job pattern the issue describes (SH12A `generatemc`-style: small consecutive `-N` values, large `nstat` per job) — jobs correlate almost completely instead of merging into independent statistics.

- Fold `-N`/`RNDOFFSET` additively into the run seed (`seed = rndseed + rndoffset`) instead of the history index, so distinct offsets decorrelate the *entire* `[0, nstat)` range.
- Remove `hist_base` from `struct osh_rng_seeding` entirely rather than leave it permanently zero — worker/replica/future-MPI splitting already runs through the separate, orthogonal `global_prim_base` argument to `osh_beam_runtime_fill_pool_at()`, which never used `hist_base` for that purpose.
- Update field docs (`osh_rng.h`, `osh_transport.h`, `osh_worker_context.h`, `osh_beam_runtime.h`) and dev docs (`docs/dev/random_numbers.md`, `src/random/README.md`, `src/transport/README.md`) to match.

## Test plan

- [x] `ctest --test-dir build_debug` — all 141 tests pass (13 pre-existing SH12A-reference skips, unrelated)
- [x] Updated `test_osh_beam_runtime_fill.c`'s overflow-guard test for the removed `hist_base` field
- [x] New regression test `test_seed_history_independent_seedoffsets` (`test_osh_rng.c`) — locks that two adjacent `RNDOFFSET`s decorrelate streams across the *whole* shared history-index range, not just a non-overlapping tail
- [x] New regression test `test_seedoffset_selects_independent_stream_family` (`test_osh_simulation.c`) — full-pipeline lock for the exact CLI scenario in the issue: `-N 0` reproducible on rerun, `-N 1` diverges
- [x] Manually re-ran the issue's repro after the fix: `-N 0` vs `-N 1` now differ in ~6500 bytes of scored payload (was 0 before the fix), matching the magnitude of `-N 0` vs `-N 2000`
- [ ] `./tools/clang-format-all.sh` / clang-tidy — not runnable in this environment (tool not installed); please confirm via CI

### DIFF
--- a/docs/dev/random_numbers.md
+++ b/docs/dev/random_numbers.md
@@ -121,10 +121,16 @@ Appropriate for the small λ of nuclear secondary multiplicities.
   into SplitMix64 seeding. Same seed + different stream ⇒ statistically
   independent sequences.
 
-Users set these through the `RNDSEED` and `RNDOFFSET` beam-file cards.
-`RNDOFFSET` is the **global history-index base** (see §6): giving each process
-or MPI rank a disjoint offset range produces disjoint, non-overlapping streams,
-so independent runs merge without correlation.
+Users set these through the `RNDSEED` and `RNDOFFSET` beam-file cards (the
+latter also settable via `-N`/`--seedoffset`). `RNDOFFSET` folds additively
+into the run seed (`seed = RNDSEED + RNDOFFSET`) to select an **independent
+stream family**: two runs sharing `RNDSEED` but using different `RNDOFFSET`
+values get statistically independent streams over the *same* history-index
+range `[0, nstat)`, so parallel array-job replicas (e.g. the SH12A
+`generatemc` convention of consecutive small `-N` values) merge without
+correlation (issue #317). This is a different axis from process/MPI/worker
+splitting (see §6), which instead assigns each worker a **disjoint
+history-index range** — `RNDOFFSET` does not touch the history index at all.
 
 ---
 
@@ -171,11 +177,11 @@ stopped.  Instead, each primary history starts its own RNG stream from its
 global history index:
 
 ```text
-history_index = RNDOFFSET + worker_range_start + worker_local_index
+history_index = worker_range_start + worker_local_index
 ```
 
-So worker 1's first primary is seeded from history index `RNDOFFSET + 250`,
-regardless of how many random numbers worker 0 consumed.  The important
+So worker 1's first primary is seeded from history index `250`, regardless of
+how many random numbers worker 0 consumed.  The important
 parallelism rule is therefore simple: worker ranges must be disjoint and
 together cover `[0, nstat)`.  Then each history ID is generated once, and each
 history gets the same random stream no matter which worker runs it.
@@ -249,7 +255,7 @@ Every history owns an **independent stream keyed by its global index**, carried
 particle, not the schedule:
 
 - **Primaries** seed from their global history index:
-  `stream_id = mix(rndseed, RNDOFFSET + prim_idx, purpose)`. A separate `BEAM`
+  `stream_id = mix(rndseed + RNDOFFSET, prim_idx, purpose)`. A separate `BEAM`
   purpose seeds a transient generator for source sampling, so the source phase
   space is identical regardless of which physics options are enabled.
 - **Secondaries** (nuclear recoils, abrasion nucleons, Fermi break-up fragments)
@@ -466,7 +472,9 @@ through TestU01 if you mean to use it for real work.
 - Per-history seeding (`osh_rng_seed_history`, `osh_rng_split`) with independent
   `BEAM`/`PHYSICS` sub-streams and lineage-deterministic secondaries.
 - One RNG stream carried per particle-pool slot, surviving compaction.
-- `RNDOFFSET` as the global history-index base for process/MPI splitting.
+- `RNDOFFSET` folded into the run seed to select an independent stream family
+  across whole runs (parallel array-job replicas); process/MPI/worker
+  splitting is the separate, orthogonal disjoint-history-range mechanism.
 - Scored-output invariance across pool capacities, up to floating-point
   reduction order (`bench::capacity_invariance`).
 
--- a/docs/dev/random_numbers.md
+++ b/docs/dev/random_numbers.md
@@ -121,16 +121,17 @@ Appropriate for the small λ of nuclear secondary multiplicities.
   into SplitMix64 seeding. Same seed + different stream ⇒ statistically
   independent sequences.
 
-Users set these through the `RNDSEED` and `RNDOFFSET` beam-file cards (the
-latter also settable via `-N`/`--seedoffset`). `RNDOFFSET` folds additively
-into the run seed (`seed = RNDSEED + RNDOFFSET`) to select an **independent
-stream family**: two runs sharing `RNDSEED` but using different `RNDOFFSET`
-values get statistically independent streams over the *same* history-index
-range `[0, nstat)`, so parallel array-job replicas (e.g. the SH12A
-`generatemc` convention of consecutive small `-N` values) merge without
-correlation (issue #317). This is a different axis from process/MPI/worker
-splitting (see §6), which instead assigns each worker a **disjoint
-history-index range** — `RNDOFFSET` does not touch the history index at all.
+Users set `seed` through the `RNDSEED` beam-file card; `RNDOFFSET` has no
+beam-file card and is set only via the `-N`/`--seedoffset` CLI flag.
+`RNDOFFSET` folds additively into the run seed (`seed = RNDSEED + RNDOFFSET`)
+to select an **independent stream family**: two runs sharing `RNDSEED` but
+using different `RNDOFFSET` values get statistically independent streams over
+the *same* history-index range `[0, nstat)`, so parallel array-job replicas
+(e.g. the SH12A `generatemc` convention of consecutive small `-N` values)
+merge without correlation (issue #317). This is a different axis from
+process/MPI/worker splitting (see §6), which instead assigns each worker a
+**disjoint history-index range** — `RNDOFFSET` does not touch the history
+index at all.
 
 ---
 
--- a/docs/dev/random_numbers.md
+++ b/docs/dev/random_numbers.md
@@ -123,15 +123,21 @@ Appropriate for the small λ of nuclear secondary multiplicities.
 
 Users set `seed` through the `RNDSEED` beam-file card; `RNDOFFSET` has no
 beam-file card and is set only via the `-N`/`--seedoffset` CLI flag.
-`RNDOFFSET` folds additively into the run seed (`seed = RNDSEED + RNDOFFSET`)
-to select an **independent stream family**: two runs sharing `RNDSEED` but
-using different `RNDOFFSET` values get statistically independent streams over
-the *same* history-index range `[0, nstat)`, so parallel array-job replicas
-(e.g. the SH12A `generatemc` convention of consecutive small `-N` values)
-merge without correlation (issue #317). This is a different axis from
-process/MPI/worker splitting (see §6), which instead assigns each worker a
-**disjoint history-index range** — `RNDOFFSET` does not touch the history
-index at all.
+`RNDOFFSET` selects an **independent stream family**: `osh_rng_seeding_init()`
+hashes `RNDSEED` and `RNDOFFSET` together through the same SplitMix64-style
+mixer used for per-history streams — `seed = RNDSEED` unchanged when
+`RNDOFFSET` is 0, otherwise `seed = mix(RNDSEED, RNDOFFSET)` — **not** a plain
+sum. Two runs sharing `RNDSEED` but using different `RNDOFFSET` values get
+statistically independent streams over the *same* history-index range
+`[0, nstat)`, so parallel array-job replicas (e.g. the SH12A `generatemc`
+convention of consecutive small `-N` values) merge without correlation (issue
+#317). A plain sum would instead let two *different* configurations collide —
+`(RNDSEED=S, RNDOFFSET=k)` and `(RNDSEED=S+k, RNDOFFSET=0)` would fold to the
+same seed and silently produce byte-identical output — so the hash-mix keeps
+that down to an ordinary, negligible 64-bit hash coincidence instead of a
+guaranteed one. This is a different axis from process/MPI/worker splitting
+(see §6), which instead assigns each worker a **disjoint history-index
+range** — `RNDOFFSET` does not touch the history index at all.
 
 ---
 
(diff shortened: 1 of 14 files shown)


grzanka:
## Critical review

I checked this out, built it (`debug` + `asan`), reran the whole suite, and reran the issue's repro on both `origin/main` and this branch. **The diagnosis and the direction of the fix hold up — I reproduced the bug and the improvement quantitatively.** But there is one blocking test defect that will turn CI red deterministically, the two new "regression locks" do not actually lock the regression, and the additive seed fold introduces a new (demonstrable) collision.

### What I could confirm

| measurement (`nstat=2000`, identical `--outdir` so the embedded path is constant) | pre-fix (`main`) | post-fix (this PR) |
|---|---|---|
| `00_minimal`, per-bin RMS relative diff of the Energy column, `-N 0` vs `-N 1` | **0.024 %** (max 0.20 %) | **15.6 %** |
| `00_minimal`, same, `-N 0` vs `-N 2000` (disjoint control) | 3.1 % (max 30 %) | 5.7 % |
| `11_plan02_geoD_mono`, `NB_target.bdo` bytes differing, `-N 0` vs `-N 1` | **2** (both timestamp digits → payload bit-identical) | 103 |
| `11_plan02_geoD_mono`, `NB_Z_narrow_dose.bdo`, `-N 0` vs `-N 1` | 2 | **6478** |

So: pre-fix, `-N 0` and `-N 1` really are the same run up to a one-history window shift (0.024 % ≈ 1/`nstat`, two orders of magnitude below the 3.1 % of a genuinely disjoint range), and post-fix adjacent offsets decorrelate at independent-run level. The "~6500 bytes" figure in the description reproduces exactly.

Two supporting checks:
- `-N 0` output is **bit-identical** pre/post fix (numeric payload of `00_minimal` unchanged), so the default path is untouched — good.
- The simplified overflow guard in `osh_beam_runtime_fill_pool_at()` is correct now that `hist_base` is gone, and `include/openshieldhit/beam.h:172` already documented `rndoffset` as *"RNG stream offset for producing independent parallel runs from the same seed"* — i.e. the old code contradicted its own public contract, and this PR makes them agree.

One caveat on the evidence as presented: on `plan02`, `NB_target.bdo` is also insensitive to `-n 2001` vs `-n 2000` (2 differing bytes), so "byte-identical" on that file is a weaker signal than it looks. The `00_minimal` 0.024 %-vs-3.1 % contrast is the sharper statement of the same bug and is worth putting in the commit message instead.

### Blocking: the new pipeline test fails 100 % of the time under the `asan` preset

`test_seedoffset_selects_independent_stream_family` compares whole TEXT payloads with `strcmp`, but the ASCII writer emits a wall-clock header line — `# Calculated %s` (`src/scoring/save/osh_scoring_save_ascii.c:111`, second resolution). The reproducibility assertion at `tests/unit/test_osh_simulation.c:1098` therefore only holds when both `-N 0` runs happen to save inside the same second:

```
build_debug: ./bin/test_osh_simulation   x12  ->  2 failures  (17 %)
build_asan:  ASAN_OPTIONS=detect_leaks=0:abort_on_error=1
             ./bin/test_osh_simulation   x8   ->  8 failures  (100 %)

ASSERT_TRUE failed: strcmp(content_0a, content_0b) == 0 (tests/unit/test_osh_simulation.c:1098)
```

Per-case runtime is ~0.25 s in `debug` and ~0.85 s under ASan, which is exactly why it is a coin flip in one and a certainty in the other. CI's `sanitizers` job runs `ctest --test-dir build_asan`, so this merges as a red pipeline; the single green `debug` run in the test plan was ~83 % luck. Fix: skip `#` comment lines (or compare parsed per-bin values) rather than raw file bytes.

### Neither new test fails on the pre-fix code

I took `origin/main` and dropped in **only** the two new test files, unmodified — they compile as-is and both pass:

```
unit::test_osh_rng .......... Passed
unit::test_osh_simulation ... Passed        # against unfixed production code
```

- `test_seed_history_independent_seedoffsets` (`test_osh_rng.c:176-185`) hand-computes `seed_a = 42 + 0`, `seed_b = 42 + 1` and asserts `osh_rng_seed_history` yields different PCG state. That is a property of `rng_mix_stream`, which held before the fix too; the test never touches `params.rndoffset` or the fold, so reverting `osh_transport_ion.c:154` leaves it green. It documents the intent but cannot detect a regression.
- The pipeline test asserts only `content_1 != content_0a`. Pre-fix, `-N 1` shifted the window to `[1, nstat]`, so one history in `nstat` differed and the payload already differed — the assertion was satisfied by the buggy code.

The property that actually broke is an *equality* the old code satisfied exactly: `stream(rndseed, N=1, h) == stream(rndseed, N=0, h+1)` for every `h`. Asserting that this equality no longer holds is a real lock and is cheap. To make it a lock on *production* behaviour rather than on arithmetic the test performs itself, the fold needs to be reachable — see the next point. Failing that, make the pipeline assertion discriminating on **magnitude** (pre-fix ≈ 1/`nstat`; post-fix tens of percent) rather than on mere inequality.

### The additive fold creates an exact alias with RNDSEED

`seed = rndseed + rndoffset` means `(RNDSEED=S, -N k)` is the *same run* as `(RNDSEED=S+k, -N 0)`. Verified on this branch:

```
(RNDSEED=89736501, -N 1)  vs  (RNDSEED=89736502, -N 0)
  -> BDO differs in 2 bytes (timestamp only); numeric TEXT payload IDENTICAL
```

With `-N` capped at 9999 and RNDSEED values in practice often small or consecutive, two campaigns that bump the seed and reuse the same `-N` range silently share runs. Merging those double-counts histories and biases the merged uncertainty low — the same class of failure #317 is about, moved onto another axis. `rng_mix_stream()` is right there and already the project's answer to this: `seeding.seed = rng_mix_stream(rndseed, rndoffset, …)`, or carry `rndoffset` as a dedicated third axis in the mixer, avoids the collision at zero cost and keeps RNDSEED the sole seed value (which also keeps `beam.h:172`'s "same seed gives bit-for-bit reproducible results" literally true).

### The fold lives in a caller, not in the module that owns the invariant

`struct osh_rng_seeding::seed` is now documented as "already folds in RNDSEED + RNDOFFSET", but the only place that does the folding is `osh_transport_ion.c:154`. Every future builder of a seeding context — the MPI path, the PHSP fill sketched in `fill_from_phsp`, a neutron-family split — has to remember to add `rndoffset` by hand, and forgetting it silently drops `-N`: precisely the bug being fixed here, and invisible to both new tests. A small constructor in `src/random/` (e.g. `osh_rng_seeding_init(s, type, rndseed, rndoffset)`) would put the semantics where they belong, keep the two user-set values traceable, and give the regression lock something production-owned to assert against.

### `-N` was the only knob for disjoint history ranges, and nothing replaces it

After this PR no CLI option can move the history-index base, yet `docs/dev/random_numbers.md` §6 and `docs/dev/scoring.md` §7 still present disjoint history ranges as the multi-node mechanism. The alternative fix — `hist_base = rndoffset * nstat` — keeps the stronger invariant (one history index ↔ one stream *globally*, so a merged campaign is provably one long run, and duplicate work is impossible rather than merely unlikely), at the cost of requiring a uniform per-job `nstat`. Seed-family folding is the more robust choice for heterogeneous jobs, so I am not arguing against it — but the trade-off should be stated in the commit message, and if a range-splitting flag is merely deferred, a TODO line would keep the dev docs honest.

### Smaller items

- `docs/user/command-line.md:76` is the only user-facing documentation of `-N` and still reads "Random seed offset override (max 9999)". This PR changes the user-visible contract of the flag, and the HPC recipe (same `RNDSEED`, distinct `-N` per job, merge weighted by `nstat`) belongs there, not only in `docs/dev/`.
- `docs/dev/bug-hunts/2026-07-05-deep-audit.md:358` still states `seed = f(rndseed, rndoffset + global index)`. Fine to leave as a dated audit snapshot — flagging in case those are treated as living docs.
- The BDO carries no `(rndseed, rndoffset)` provenance, so a merge tool cannot detect the duplicate-replica case above. Worth an issue if not already tracked.
- Test plan, last box: clang-format **is** runnable — I ran clang-format 18 over all nine changed files. The only report is `src/beam/runtime/osh_beam_runtime.c:42` (`rt->primary = (struct particle) {0};`), which is present on `main` too and outside this diff. The PR is format-clean; that box can be ticked.
- `ctest --test-dir build_debug`: 141/141 pass (13 `reference::` skips), matching the description.

### Repro

```bash
# flake, plain debug build
cmake --preset debug && cmake --build --preset debug --parallel
for i in $(seq 12); do ./build_debug/bin/test_osh_simulation >/dev/null || echo "fail $i"; done

# same test under the preset the CI sanitizers job uses
cmake --preset asan && cmake --build --preset asan --parallel
ASAN_OPTIONS=detect_leaks=0:abort_on_error=1 ./build_asan/bin/test_osh_simulation

# tests do not lock the regression: main + only the two new test files
git checkout origin/main && git checkout <this-branch> -- tests/unit/test_osh_rng.c tests/unit/test_osh_simulation.c
cmake --build --preset debug --parallel && ctest --test-dir build_debug -R "test_osh_rng|test_osh_simulation"

# seed aliasing
sed -i 's/89736501/89736502/' tests/cases/00_minimal/beam.dat   # then compare against -N 1 at the original seed
./build_debug/bin/openshieldhit -n 2000 -N 0 --outdir out tests/cases/00_minimal/
```

**Summary:** fix the timestamp comparison (blocking, CI-red as-is), make at least one test fail on pre-fix code, and please reconsider `+` in favour of `rng_mix_stream` for the fold. The rest is documentation and follow-up scope.

---

grzanka:
Pushed two commits addressing the review above:

**`5228656` random: hash-combine RNDOFFSET into the seed instead of adding it**
- New `osh_rng_seeding_init(seeding, type, rndseed, rndoffset)` in `src/random/` centralizes the fold (previously inline at the one transport call site). `rndoffset == 0` stays an identity (`seed == rndseed`), so the default single-run path is bit-for-bit unchanged — verified against both this PR's pre-fix output and `main`.
- The fold now hashes `rndseed`/`rndoffset` through the existing `rng_mix_stream()` mixer instead of `+`. Addition let `(RNDSEED=S, RNDOFFSET=k)` alias `(RNDSEED=S+k, RNDOFFSET=0)` — verified locally, byte-identical payload — which would have silently double-counted histories if two such campaigns were ever merged as independent replicas. Deliberately **not** doing the alternative (`hist_base = rndoffset * nstat`) here: per-job `nstat` doesn't need to be uniform, so a seed-family hash is the right fit rather than a range multiply.
- `test_seed_history_independent_seedoffsets` now goes through the real `osh_rng_seeding_init()` instead of hand-computing `rndseed+offset`; added `test_seedoffset_does_not_alias_rndseed` as a direct lock on the new failure mode.
- Documented the `-N`/`--seedoffset` array-job pattern in `docs/user/command-line.md` (previously only the terse `--help` mirror), including that replicas merge by each file's own `nstat`.

**`7b794c1` tests: strip wall-clock header before comparing seedoffset payloads**
- `test_seedoffset_selects_independent_stream_family` was `strcmp`-ing whole files including the ASCII writer's `# Calculated <timestamp>` line, so it flaked whenever two `-N 0` runs straddled a one-second boundary — reproduced 8/8 under the `asan` preset before this fix (slower per-run, so the boundary was hit almost every time). Now strips `#` lines before comparing.

Re-verified end to end on a clean rebuild: `ctest --test-dir build_debug` and `ctest --test-dir build_asan` (with `ASAN_OPTIONS` matching CI) both 141/141; `clang-format-18` and `clang-tidy-18` (via `clang-tidy-diff-18.py` on the `main`-diff, mirroring `.github/workflows/clang-tidy.yml`) clean on all touched files; the previously-flaky test run 20x under both presets with 0 failures.

---

